PR #11, opened 2020-03-05
Junior devs

alexb8z:
The login pages have proper wording from the designers, and the code is neater and less redundant now.

### DIFF
--- a/components/auth/RecoverPasswordFormContent.tsx
+++ b/components/auth/RecoverPasswordFormContent.tsx
@@ -22,7 +22,7 @@ const PasswordAssistanceCaption =
 const PasswordAssistanceCaption2 = "We have sent a new password to this email.";
 const ForgetPasswordCTAText = "Back to Sign In";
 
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -13,7 +13,6 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "strict": true,
```
PC98: @alexb8z Could you undo changes to this file? Let's keep the strict mode.

components/auth/RecoverPasswordFormContent.tsx
```diff
@@ -0,0 +1,87 @@
...
+    },
+    []
+  );
+
+  const onClickGoBack = useCallback(() => {
+    navigateToContent("login");
+  }, [navigateToContent]);
+
+  const onPressCTA = useCallback(() => {
+    //debugger;
```
PC98: Nit: delete commented out code

PR #15, opened 2020-03-10
Donation Page Contact Form and other UI changes

PC98:
* Adjust CSS to ensure that the donation page is scrollable for long forms.
* Change auth page component prop types to support async callback functions.
* Refactor code to use `img` tags instead of `div`s with `backgroundImage` wherever possible.
* Get started on donation page contact form.
* `MuiTextField`s are `filled` by default now. Also changed `backgroundColor` of all filled text fields using theme overrides.


utils/theme.ts
```diff
@@ -27,7 +27,8 @@ export default createMuiTheme({
   },
   props: {
     MuiTextField: {
-      variant: "filled"
+      variant: "filled",
+      size: "small"
```
PC98: The designs actually use regular sized text fields with height 56px. But I think this change makes the UI look better.

PR #16, opened 2020-03-10
Undoes certain changes in commit 99eefef

PC98:


components/auth/AuthPageFormContainer.tsx
```diff
@@ -11,7 +11,7 @@ const useStyles = makeStyles({
   container: {
     width: "34vw",
     height: "52vh",
-    minWidth: 400,
+    minWidth: 360,
```
PC98: `400px` is too big for low resolution (800 x 600) laptop devices. With this change, we can expect the UI to look like this:
Screen Shot 2020-03-10 at 1 03 33 PM

components/auth/AuthPageLayout.tsx
```diff
@@ -13,36 +13,27 @@ const useStyles = makeStyles(({ margins }: Theme) =>
       justifyContent: "space-between"
     },
     content: {
-      display: "flex",
```
PC98: styles not needed anymore

components/auth/AuthPageLayout.tsx
```diff
@@ -13,36 +13,27 @@ const useStyles = makeStyles(({ margins }: Theme) =>
...
-      flex: 0.6
-    },
-    main: {
       position: "relative",
       display: "flex",
-      minWidth: 500,
```
PC98: It seems to be better to allow the form itself to determine `content` `div`'s `minWidth`, now that the background color for the `div` matches the image's background color.

components/auth/AuthPageLayout.tsx
```diff
@@ -13,36 +13,27 @@ const useStyles = makeStyles(({ margins }: Theme) =>
...
       flexDirection: "column",
       justifyContent: "center"
     },
     placeholder: {
       height: "68vh",
-      minHeight: 450,
-      margin: "0 auto",
```
PC98: `margin: "0 auto"` will horizontally center the `img`. We don't want that - as per Figma designs, the image is right justified.

PC98: Not sure if entirely need a `minHeight` here, but I am open to talking about it

components/auth/LoginFormContent.tsx
```diff
@@ -17,15 +17,13 @@ const LoginFormContent: React.FC<ContentComponentProps> = ({
   const onChangeEmail = useCallback(
     (e: React.ChangeEvent<HTMLInputElement>) => {
       setEmail(e.target.value);
-      setHasError(false);
```
PC98: Ideally, we don't want `setHasError` to be set to `false` when the user starts typing again. Trying to emulate what Google does with its login form - if you try signing in to your Google account and enter an incorrect email, the error message doesn't disappear until you try submitting your credentials again.

components/donation/DonationPageFormBody.tsx
```diff
@@ -15,7 +15,7 @@ const useStyles = makeStyles({
     display: "flex",
     flex: 1,
     flexDirection: "column",
-    padding: "4.5vh 4vh",
+    padding: "3.5vh 3vh",
```
PC98: `3.5vh` and `3vh` are more appropriate. They map to approx. `32px` and `28px` as per Figma designs on a 1440 x 900 screen.

components/donation/DonationPageHeader.tsx
```diff
@@ -5,13 +5,14 @@ import { createStyles, makeStyles, Theme } from "@material-ui/core/styles";
 import DonationPageHeaderSelect from "./DonationPageHeaderSelect";
 import DonationPageHeaderBranding from "./DonationPageHeaderBranding";
 
+const white = "white";
```
PC98: While we don't have an eslint-rule configured for this, it is generally preferred to not use color literals. See: https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-color-literals.md

PR #24, opened 2020-04-01
Added Backend Functionalities

nichabosh:
### Description
Because I primarily implemented functionalities on the backend, a lot of the work completed isn't immediately present in the application. If you'd like to delve a bit deeper into some of the changes made, here's a quick log of the more hefty commits.

### Commits
```
commit 6e4ded28e0dd055e286f37b2643afda31eee6e97
Author: Nicholas Boshart <nichabosh.dev@gmail.com>
Date:   Mon Mar 30 03:54:45 2020 -0500

    Added getOrgNames HTTP Request
    
    • Renames 'pages/actions' directory to 'pages/requests'.
    • Adds comments to all request handlers in pages/api.
    • Includes all necessary files & functions for getOrgNames.


commit f703cc32491b990db1850176b911a60fcaaa8b44
Author: Nicholas Boshart <nichabosh.dev@gmail.com>
Date:   Sun Mar 29 23:09:58 2020 -0500

    Added Login Request & Implementation
    
    • Created admin HTTP requests in pages/actions (needs to be changed).
    • Modified error messages & made slight changes to login function.
    • Added isomorphic-unfetch to dependencies & page routes to config.js.
    • Connected login functionality to auth pages in the frontend.


commit 9c239fb5b3a88c0938dd2a64953b83fdf76d2442
Author: Nicholas Boshart <nichabosh.dev@gmail.com>
Date:   Wed Mar 11 18:59:36 2020 -0400

    Added Endpoints for Donation & Nonprofit
    
    • Ensures admin accounts cannot be made without correspondence to a nonprofit.
    • Includes a means of creating nonprofits and their related donations.
    • Adds necessary comments to endpoints for additional clarity.
    • Changes "logo" attribute of nonprofit to a string (temporary).


commit 2a83dd1dd88d3fd8439758bfd29f2006583b0722
Author: Nicholas Boshart <nichabosh.dev@gmail.com>
Date:   Sun Mar 1 20:59:21 2020 -0500

    Added Signup Functionality
    
    • Includes necessary endpoints and request handlers for signup.
    • Establishes correct access to environment variables.
    • Adds config.js in root directory for convenience.
    • Slight stylistic changes to object models.
```


requests/admin.js
```diff
@@ -0,0 +1,24 @@
+import fetch from "isomorphic-unfetch";
+import urls from "../config";
+
+export const login = async (email, password) => {
+  return fetch(urls.baseUrl + urls.apis.login, {
```
PC98, line 5: Should be able to remove `urls.baseUrl` (in reference to my previous comment)

server/actions/admin.js
```diff
@@ -1,25 +1,104 @@
...
 
 export async function login(email, password) {
   await Mongo();
 
-  return Admin.findOne({ email: email }).then(admin => {
-    if (admin) {
-      return bcrypt.compare(password, admin.password).then(result => {
-        return result
+  return Admin.findOne({ email: email })
+    .then(async admin => {
```
PC98, line 19: Could you run `yarn lint`, if you haven't already 😅

components/auth/LoginFormContent.tsx
```diff
@@ -31,8 +37,17 @@ const LoginFormContent: React.FC<ContentComponentProps> = ({
   );
 
   const onPressCTA = useCallback(async () => {
-    // TODO: Code for signing-in
-  }, []);
+    return login(email, password)
```
PC98, line 40: I don't we should be returning anything here. Also we should probably refactor to use `await` since this is an `async` function:
```
try {
  await login(email, password);
  // do stuff with router
} catch (err) {
 // do stuff with err
}
```

pages/api/createNonprofit.js
```diff
@@ -0,0 +1,22 @@
+import { createNonprofit } from "server/actions/nonprofit";
+
+// @route   POST api/createNonprofit
+// @desc    Creates an Organization
+// @access  Public
+const handler = (req, res) => {
+  createNonprofit(req.body.name, req.body.about, req.body.logo, req.body.colors)
```
PC98, line 7: Do we need a `/pages/api/createNonprofit` endpoint, or will be using it in the near future? In general we should avoid unused code since it increases our app's bundle size.

pages/api/getOrgNames.js
```diff
@@ -0,0 +1,22 @@
+import { getOrgNames } from "server/actions/nonprofit";
+
+// @route   GET api/getOrgNames
+// @desc    Requests List of Nonprofits
+// @access  Public
+const handler = (req, res) => {
```
PC98, line 6: Nit: to be consistent, we should probably use nonprofit or nonProfit instead of org in our naming scheme.

pages/api/signup.js
```diff
@@ -0,0 +1,28 @@
+import { signup } from "server/actions/admin";
+
+// @route   POST api/signup
+// @desc    Requests Admin Creation
+// @access  Public
+const handler = (req, res) => {
```
PC98, line 6: Same as `/pages/api/createNonprofit`, do we need this endpoint or will be using it in the near future?

requests/admin.js
```diff
@@ -0,0 +1,24 @@
...
+    method: "post",
+    mode: "same-origin",
+    credentials: "include",
```
PC98, line 8: Could leave out `credentials` since our request is to the same origin

requests/nonprofit.js
```diff
@@ -0,0 +1,23 @@
+import fetch from "isomorphic-unfetch";
+import urls from "../config";
+
+export const getOrgNames = async () => {
+  return fetch(urls.baseUrl + urls.apis.getOrgNames, {
+    method: "get",
+    mode: "same-origin",
+    credentials: "include",
+    headers: {
```
PC98, line 9: I don't think a content-type header doesn't makes sense for `GET` requests, so you should be able to remove this. You can also omit the  `method: "get"` since `fetch` defaults to `GET` and you can also skip `credentials: "include"`

requests/nonprofit.js
```diff
@@ -0,0 +1,23 @@
...
+      "Content-Type": "application/json"
+    }
+  })
+    .then(res => res.json())
+    .then(json => {
+      if (!json) {
```
PC98, line 15: Nit: since you are throwing, you don't really need the `else`s here

server/models/admin.js
```diff
@@ -18,19 +18,17 @@ const adminSchema = new Schema({
...
-  this.password = hash;
-  next();
-});
-
 if (process.env.IS_OFFLINE) {
   delete mongoose.connection.models.Admin;
 }
 
-module.exports = mongoose.model("Admin", adminSchema);
+module.exports = mongoose.models.Admin || mongoose.model("Admin", adminSchema);
```
PC98, line 34: Does this, and other similar changes in other files, render lines 30-32 useless? I think both achieve the same purpose?

server/actions/nonprofit.js
```diff
@@ -0,0 +1,37 @@
...
+        });
+  });
+}
+
+export async function getOrgNames() {
+  await Mongo();
+
+  // search through the 'nonprofit' collection with no filter, thus
+  // returning every document present in the database.
+  return Nonprofit.find().then(organizations => {
```
PC98, line 30: Nit:
```
return Nonprofit.find().then(organizations => organizations.map(org => org.name))
```

components/auth/LoginFormContent.tsx
```diff
@@ -64,7 +79,7 @@ const LoginFormContent: React.FC<ContentComponentProps> = ({
         password={password}
         onChangePassword={onChangePassword}
         hasError={hasError}
-        hasErrorHelperText="Incorrect email or password"
+        hasErrorHelperText={error}
```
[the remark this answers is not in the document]
nichabosh: From what I've created thus far, all error messages that will eventually be displayed to the user are generated on the backend. I intend to stick to this in the future; however, if you feel that there's a better way to approach this, feel free to let me know.

next.config.js
```diff
@@ -1,6 +1,13 @@
 const path = require("path");
+const dotenv = require("dotenv");
+
+dotenv.config();
 
 module.exports = {
+  env: {
```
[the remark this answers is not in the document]
nichabosh: Ah, good catch. I meant to delete this export after creating `config.tsx`, but it apparently slipped away from me. Thanks for letting me know!

components/auth/LoginFormContent.tsx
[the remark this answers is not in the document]
PC98: The problem is since this is a blanket `try-catch` block, you will be catching all errors thrown in dealing with this `Promise`, and not just the errors we have coded. For example, the `router` code may throw an error, and the error message generated by JavaScript/Next.js may be something cryptic and should not be display to the user as text under the password field. Further, giving
 specific error messages at failed login attempts is a security concern, you probably don't want to tell a bad actor which of the fields - email or password - is incorrect. So I think we should stick to a generic error message on the front-end, but maybe log the error for developers to get to later.

server/actions/admin.js
nichabosh: It seems to comply with the standards set by the linter. I actually like this syntax a bit better than keeping the `.then(...)` on the same line, as it (personally) makes reading through callbacks easier.

pages/api/createNonprofit.js
nichabosh: I created this endpoint in the earlier stages of the project because we initially planned on having organizations sign up via our webapp. Though this functionality isn't quite part of our MVP _this_ semester, its implementation is still up in the air for Spring, so I think it's best to leave it for now.

pages/api/getOrgNames.js
nichabosh: Yeah, I began to notice this as well. Just went back and standardized everything to be in reference to a "nonprofit".

pages/api/signup.js
nichabosh: This is actually necessary because, even if we're creating the admin accounts ourselves, we have to use libraries like `bcrypt` to sufficiently hash and salt the passwords. In the future, though, I'd like to move toward having an independent page in which admins would be able to sign-up and request access to their nonprofit.

requests/nonprofit.js
nichabosh: Ah, good catch—forgot to remove `header` after deleting `body`. I'm going to go ahead and leave `method: "get"` in the code because, though it may not be necessary, I feel that it provides good clarification.

pages/api/signup.js
PC98: So if we are creating the admin accounts ourself, we would do so via a simple JS script which directly makes transactions to our DB. It would basically be a DB seed script.

I would recommend not exposing the sign-up functionality through an API for now. There could be security issues - for example, anyone can make a `POST` request to this endpoint and claim that they are an admin for a  particular non-profit.

components/auth/LoginFormContent.tsx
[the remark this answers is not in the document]
nichabosh: Okay, that makes sense. I went ahead and simulated what it'd be like to receive an error from an unexpected origin and found that generic feedback such as "Invalid email or password" will always be displayed, regardless of whether it's true or not (i.e. `router` errors will produce the same message). I couldn't quite figure out a good way to get around this, so I took the error messages out of each individual file in `server/actions` and instead put them into a collective file,`utils/errors.js`, which can be accessed anywhere in our application. Now, I'm conditionally displaying either the generic error message _or_ "An unexpected error occurred" depending on the type of error present, if that makes sense. Go ahead and look through the updated code & let me know if you find anything wrong with this implementation.

pages/api/signup.js
nichabosh: I understand the security concern, but I'd rather leave it for the time being. I'll make a comment in the file & we can revisit the issue when we're closer to releasing a full-fledged application.

PR #31, opened 2020-04-06
Admin login link

alexb8z:
I think I have pretty much finished up the admin login access here by following Michelle's design advice, but it is still very flexible and can easily be changed if needed.

### DIFF
--- a/components/AdminLoginLink.tsx
+++ b/components/AdminLoginLink.tsx
@@ -29,17 +29,12 @@ const AdminLoginLink: React.FC<React.ComponentProps<typeof Link>> = ({
   const { container, text } = useStyles();
 
   return (
-    <div className={clsx(container)}>
-      <Typography
-        className={clsx(text)}
-        color={color}
-        {...rest}
-        variant="subtitle2"
-      >
+    <div className={container}>
+      <Typography className={text} color={color} {...rest} variant="subtitle2">
         Are you an admin?
       </Typography>
       <Link
-        className={clsx(text)}
+        className={text}
         color={color}
         {...rest}
         href="/login"
--- a/components/AdminLoginLink.tsx
+++ b/components/AdminLoginLink.tsx
@@ -2,35 +2,51 @@ import React from "react";
 
 import clsx from "clsx";
 
-import { Theme, makeStyles, createStyles } from "@material-ui/core/styles";
+import makeStyles from "@material-ui/core/styles/makeStyles";
 
 import Link from "@material-ui/core/Link";
 
 import Typography from "@material-ui/core/Typography";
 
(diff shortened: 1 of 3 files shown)


components/AdminLoginLink.tsx
```diff
@@ -0,0 +1,55 @@
...
+const AdminLoginLink: React.FC<React.ComponentProps<typeof Link>> = ({
+  children,
+  className,
+  color = "secondary",
+  ...rest
+}) => {
+  const { container, text } = useStyles();
+
+  return (
+    <div className={clsx(container)}>
```
PC98, line 32: **Nit:** don't use `clsx` if `className` is just one string/single element, so something like: `className={container}`  should suffice.

components/AdminLoginLink.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <Typography
+        className={clsx(text)}
+        color={color}
+        {...rest}
+        variant="subtitle2"
+      >
+        Are you an admin?
+      </Typography>
+      <Link
```
PC98, line 41: Could you instead try using the `Link` component defined in `components/Link`? It servers as a material-ui wrapper around Next.js's `Link` and should support all these props.

components/AdminLoginLink.tsx
alexb8z: fixing it and recommiting!

components/donation/DonationPageFormBody.tsx
```diff
@@ -111,6 +112,7 @@ const DonationPageFormBody: React.FC = () => {
         >
           Continue
         </ButtonWithNonprofitColor>
+        <AdminLoginLink />
```
PC98, line 115: I would say since we are not supplying any props to `AdminLoginLink`, you can omit it's prop types in the component's definition: 
```
const AdminLoginLink: React.FC = () => {
    // stuff
}
```

components/AdminLoginLink.tsx
```diff
@@ -0,0 +1,50 @@
+import React from "react";
+
+import clsx from "clsx";
```
PC98, line 3: Feel free to remove this import when you remove `clsx` usage

alexb8z:
> Alex, could you post a screenshot of what the updated UI looks like?

Screen Shot 2020-04-05 at 9 33 45 PM

components/AdminLoginLink.tsx
alexb8z: Removed the import and the clsx usage

components/donation/DonationPageFormBody.tsx
alexb8z: removed the unnecessary props

PR #32, opened 2020-04-06
Backend refactors and a couple of bug fixes

PC98:
In this PR:
* Added `process.env.JWT_SECRET` to `config.ts` - now all environment variables are defined there.
* Converted all `pages/api` code to TypeScript, and created a utility function to DRYup code.
* Remove unnecessary `tsconfig.json` options.
* Bug fix: Hitting any of our async API endpoints would print the following warning from Next.js in the terminal: `API resolved without sending a response for /api/some_endpoint, this may result in stalled requests.` This has been fixed.
* User wouldn't be redirected to `/` due to the usage of `getInitialProps` in `IndexPage`. This bug has also been fixed.
* Cleaned-up backend code and refactored it to use `async-await` 
* The submit button on the login page stays in the loading state until the page transition to `/` fully completes.


components/donation/reducer.ts
```diff
@@ -0,0 +1,90 @@
...
+  curStepIndex: number;
+  isContinueButtonDisabled: boolean;
+  contactStep: ContactStepProps;
+  amountStep: AmountStepProps;
+};
+
+export const AMOUNTS = [25, 50, 100, 250, 500];
+
+export const MIN_OTHER_AMOUNT = "0";
+export const MAX_OTHER_AMOUNT = "2000";
```
nichabosh, line 26: **NIT:** I probably should've pointed this out earlier—and it's not a big deal at the moment—but couldn't people abuse the minimum donation amount by setting a value such as $0.01? I personally think it should be raised to around $2.50 or $5.00, but I suppose it's more of a design choice.

PC98: Yep this is something I discussed with Chima and the designers. They had said > $0 should be the min but I do think we should change that since I think Stripe will charge a processing fee. I will float this issue on our Slack.

PC98:
* I also thought that an explicit `Back` button would be better/more obvious, but designers told me to use navigation bar, at least for now. I can ask them again though.
* I do agree that navigation in such a manner should be allowed, but decided against implementing it since it since it would make our state a bit complicated. But I can implement this behavior in a future PR.

PR #36, opened 2020-04-14
Using Stripe Elements for last Donation Form

PC98:
More PRs to come


components/donation/DonationPageFormBody.tsx
```diff
@@ -26,7 +37,10 @@ const useStyles = makeStyles({
     backgroundColor: white
   },
   contentContainer: {
-    flex: 3
+    flex: 5,
+    marginTop: 8,
+    marginBottom: 8,
+    minHeight: 260
```
PC98, line 43: This is a bit hacky currently but intentional. Problem is that the last donation form is really big so it may trigger a display height increase on certain screen sizes when it renders. When that happens, the background image sort of flickers, due to the `backgroundSize: "cover"` property.

I am not that good with CSS to figure out a solution to this problem for all screen sizes, but setting this `minHeight` should help.

nichabosh: I, too, am not a CSS expert. Though it's not an end-all be-all solution, I think the `minHeight` route should be fine—at least temporarily. It seems to work well enough for "standard" screen sizes, but we can revisit down the road if necessary.

PR #37, opened 2020-04-14
MVP

PC98:


components/donation/DonationPageFormBody.tsx
```diff
@@ -121,16 +138,25 @@ const DonationPageFormBody: React.FC<Props> = ({
           setIsSubmitting(true);
           // TODO: Check if we should use paymentStep.name instead
           const name = `${contactStep.firstName} ${contactStep.lastName}`;
```
PC98, line 140: @nichabosh Let me know what you think of this `TODO`. Basically, currently I am passing the name collected from the "Contact" step to both Stripe and our own createDonation API endpoint. I am not sure if I should be passing the name collected in the "Payment" step instead though.

nichabosh: I think it makes sense to use the name in the "Contact" section for donation-logging purposes, whereas we should pass the name used in "Payment" to Stripe. If someone were to donate using a card in someone else's name—say, a kid using their parents'—it would make sense for the donation to be in reference to the kid. When passing the payment information to Stripe, however, I think it should include the actual payer's name for record-keeping. It's not a super pressing issue, though, and, if Stripe needed the actual card holder's name, then they likely would have required it by default.

PR #61, opened 2020-08-26
More clean-up

PC98:
* Removed some "risky" endpoints. We should not let non-profit admins sign-up via a publically accessible URL -- anyone can hit this endpoint and become an admin of a non-profit in our DB! Similarly, also removing the create non-profit endpoint. I don't think we should, at least for now, have a public, unsecured endpoint for this. **Note that I am not deleting server-side code for either of these endpoints, so should we choose to add them back, it will be straightforward.**
* Some other cleanups, like using `.exists` instead of `.findOne`.


components/auth/form/LoginFormContent.tsx
```diff
@@ -37,9 +37,8 @@ const LoginFormContent: React.FC<ContentComponentProps> = ({
         router.push(urls.pages.index);
       } catch (err) {
         setError(
-          err.message === errors.admin.INVALID_EMAIL ||
-            err.message === errors.admin.INVALID_PASSWORD
-            ? "Incorrect email or password."
+          err.message === errors.admin.INVALID_LOGIN
```
PC98, line 40: For safety concerns, we should not specifically expose to our client-side code if it was the email that was invalid or if it was the password. So we replace `errors.admin.INVALID_EMAIL` and `errors.admin.INVALID_PASSWORD` with a more generic `errors.admin.INVALID_LOGIN`

PR #65, opened 2020-08-26
Upgrade to Next.js v9.5, TypeScript v4.0

PC98:
* Ran `yarn upgrade`.
* Updated to using Next.js v9.5, TypeScript v4.0 (and latest React and TS ESLint versions).
* Migrating to TypeScript v4.0 caused some new lint errors, so I also fixed those.
* **Next.js v9.5 changes how we handle env variables, details here: https://nextjs.org/docs/basic-features/environment-variables**. TL;DR:
    * We will store prod secrets and env variables in `.env.local` file.
    *  `.env.development` will contain dev env variables.
    * `.env.development.local` will contain dev secrets.
    * `.gitignore` will ensure that `.env.development.local` and `.env.local` are not tracked by git. **However, `.env` file is no longer a part of our `.gitignore`.** This is intentional in Next.js v9.5 since `.env` is no longer supposed to store secrets.
* Bitwarden and `yarn secrets` scripts have been modified so that other devs can sync `.env.development.local` and `.env.local`.
* Updated `.eslintrc.json` to work with latest version of `@typescript-eslint/parser`. We will be using `babel-eslint` parser for `*.js` and `*.jsx` files now.
* Modified `.gitignore` to match Next.js v9.5 default's `.gitignore`.
* Removed `next.config.js` -- it's no longer needed.

PR #66, opened 2020-08-27
Sync secrets using bitwarden item ID

Arthelon:
Also switches to using the development `.env` file instead of the production one


Arthelon:
Still not sure whether using Bitwarden item names or IDs would be the best. Going to close this for now.

PC98:
Okay. @Arthelon feel free to suggest changes in #65, it makes changes to the `yarn secrets` script and modifies how env variables and secrets are managed for gen-soln.

PR #76, opened 2020-08-29
Support MongoDB migrations

PC98:
* Using [`migrate-mongo`](https://github.com/seppevs/migrate-mongo)'s APIs, we can add migration support to MongoDB.
* Migration scripts will help us seed our database and propagate schema changes from dev DBs to prod DB, and one person's dev DB to another person's dev DB. More details on migrations: https://softwareontheroad.com/database-migration-node-mongo/
* Going forward, any change to our schema (like addition/deletion of a field) should be accompanied by a migration file.
* Creating a migration:
```
yarn dev:db:migrate create some-description-of-migration
```
* Applying **all** migrations to your dev DB:
```
yarn dev:db:migrate up
```
* Roll-back the most recent migration:
```
yarn dev:db:migrate down
```
* View migration status:
```
yarn dev:db:migrate status
```
* The migration script imports `"config.ts"`, so we re-use all DB related configurations (like DB name, DB URL) defined in that file.


server/migrations/20200829032827-seed-nonprofits.ts
```diff
@@ -0,0 +1,48 @@
+import { Db } from "mongodb";
```
KoolCards, line 1: Should we be pushing the seed files created by the migrations script?

PC98: Yeah, since this is dummy data and we want to make sure that everyone's localhost DB contains some data at least so that they can render a few non-profit donate pages.

Issue #79, opened 2020-08-30
[DMS] Rename `createDonation` API endpoint and server action.


PC98:
* After Stripe successfully processes a donation, we log the details in our DB by creating a Donation object.
* The name `createDonation` is a bit misleading -- this API endpoint (and the corresponding server action) doesn't process the payment or interact with Stripe, it simply creates the Donation object for logging purposes.
* Let's rename the `/api/createDonation` API endpoint to `/api/logDonation` so that this more clear. A very similar task is #77, look at steps outlined there for additional help.
* Additionally, let's also rename the server action (see `server/actions/donation.ts`) and the client request (see `requests/donation.ts`) to `logDonation`. TL;DR every occurrence of `createDonation` in the codebase should be renamed to `logDonation`.
* Test your change! Make a dummy donation using a sample credit card from here: https://stripe.com/docs/payments/accept-a-payment#web-test-integration, then open up the [MongoDB shell](https://docs.mongodb.com/manual/mongo/) locally to check if the expected donation was logged.

PR #80, opened 2020-08-30
Bunch of changes/improvements

PC98:
See comments in the PR for details

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -15,6 +15,7 @@
     "plugin:prettier/recommended"
   ],
   "rules": {
+    "no-return-await": "error",
     "react/prop-types": "off",
     "react/display-name": "off",
     "react/react-in-jsx-scope": "off",
@@ -47,6 +48,7 @@
         "plugin:prettier/recommended"
       ],
       "rules": {
+        "no-return-await": "error",
         "react/prop-types": "off",
         "react/display-name": "off",
         "react/react-in-jsx-scope": "off",
--- a/components/donation/form/DonationPageForm.tsx
+++ b/components/donation/form/DonationPageForm.tsx
@@ -107,32 +107,34 @@ const DonationPageForm: React.FC<Props> = ({
       if (!e.currentTarget.reportValidity()) return;
 
       if (isLastStep) {
+        setIsSubmitting(true);
+        // TODO: Check if we should use paymentStep.name instead
+        const name = `${contactStep.firstName} ${contactStep.lastName}`;
+        const email = contactStep.email;
+
         try {
-          setIsSubmitting(true);
-          // TODO: Check if we should use paymentStep.name instead
-          const name = `${contactStep.firstName} ${contactStep.lastName}`;
-          const email = contactStep.email;
+          // Deliberately limit this try-catch only to Stripe's payment processing.
+          // Catching errors from other miscellaneous code might make it seem like the payment failed, even when it succeeded.
           await processPayment(
             name,
             contactStep.email,
             paymentStep.zipcode,
             amount
           );
-          // TODO: What should we do if Stripe has processed the payment correctly, but our createDonation API call errored?
-          await createDonation({
-            name,
-            email,
-            amount,
-            nonprofitId: selectedNonprofitId
-          });
-
-          donationCompletedCallback();
-          // Don't call setIsSubmitting(false) after this -- donationCompletedCallback() will unmount this component
         } catch (err) {
           // TODO: Not sure how else to handle and display the error
           setIsSubmitting(false);
           err instanceof Error && alert(err.message);
+          return;
         }
+        donationCompletedCallback(); // Don't call setIsSubmitting(false) after this -- donationCompletedCallback() will unmount this component
+        // TODO: What should we do if Stripe has processed the payment correctly, but our createDonation API call errored?
+        void createDonation({
+          name,
+          email,
+          amount,
+          nonprofitId: selectedNonprofitId
+        });
       } else {
         dispatch(incrementStep());
       }
--- a/components/donation/form/DonationPageForm.tsx
+++ b/components/donation/form/DonationPageForm.tsx
@@ -1,4 +1,11 @@
-import React, { useCallback, useMemo, useState, useReducer } from "react";
+import React, {
+  useCallback,
+  useMemo,
+  useState,
+  useReducer,
+  useEffect
+} from "react";
+import { Router } from "next/router";
 import dynamic from "next/dynamic";
 
 import makeStyles from "@material-ui/core/styles/makeStyles";
@@ -72,17 +79,12 @@ const DonationPageForm: React.FC<Props> = ({
 }) => {
   const { container, contentContainer } = useStyles();
   const [
-    {
-      curStepIndex,
-      isContinueButtonDisabled,
-      contactStep,
-      amountStep,
-      paymentStep
-    },
+    { curStepIndex, isCurStepCompleted, contactStep, amountStep, paymentStep },
     dispatch
   ] = useReducer(reducer, initialState);
 
   const [isSubmitting, setIsSubmitting] = useState(false);
+  const [isRouteChanging, setIsRouteChanging] = useState(false);
 
   const { isReady, processPayment } = useStripePayment();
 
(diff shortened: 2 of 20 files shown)


.eslintrc.json
```diff
@@ -15,6 +15,7 @@
     "plugin:prettier/recommended"
   ],
   "rules": {
+    "no-return-await": "error",
```
PC98, line 18: New lint rule that we will error on. Prevents people from using `return await something` in `async` functions since that degrades performance.

components/donation/form/DonationPageForm.tsx
```diff
@@ -72,17 +79,12 @@ const DonationPageForm: React.FC<Props> = ({
...
   const { container, contentContainer } = useStyles();
   const [
-    {
-      curStepIndex,
-      isContinueButtonDisabled,
-      contactStep,
-      amountStep,
-      paymentStep
-    },
+    { curStepIndex, isCurStepCompleted, contactStep, amountStep, paymentStep },
```
PC98, line 82: Renamed `isContinueButtonDisabled` to `isCurStepCompleted`, to make it more clear that this variable just tells us if the current step in the donation form is completed or not.

components/donation/form/DonationPageForm.tsx
```diff
@@ -72,17 +79,12 @@ const DonationPageForm: React.FC<Props> = ({
...
     dispatch
   ] = useReducer(reducer, initialState);
 
   const [isSubmitting, setIsSubmitting] = useState(false);
+  const [isRouteChanging, setIsRouteChanging] = useState(false);
```
PC98, line 87: We will disable the continue button whenever the route is changing, like when the user clicks on the "Login here" link or selects another non-profit from the dropdown. I don't think the user should be allowed to submit/progress in the form when the route is changing -- might lead to weird bugs.

components/donation/form/DonationPageForm.tsx
```diff
@@ -172,6 +174,24 @@ const DonationPageForm: React.FC<Props> = ({
...
+  useEffect(() => {
+    // No need to call setIsRouteChanging(true) on routeChangeComplete since this entire component in re-mounted when that happens
+    Router.events.on("routeChangeStart", routeChangeStart);
+
+    return () => {
+      Router.events.off("routeChangeStart", routeChangeStart);
+    };
+  }, [routeChangeStart]);
+
+  const isContinueButtonDisabled = useMemo(
```
PC98, line 190: Putting all the conditions together in one `useMemo` variable.

components/donation/form/DonationPageFormAmountStep.tsx
```diff
@@ -52,16 +52,16 @@ const DonationPageFormAmountStep: React.FC<AmountStepProps> = ({
     radioButtonAmount
   ]);
 
-  const isContinueButtonDisabled = useMemo(
+  const isCurStepCompleted = useMemo(
     () =>
-      hasSelectedOther &&
-      (otherAmount < +MIN_OTHER_AMOUNT || otherAmount > +MAX_OTHER_AMOUNT),
+      !hasSelectedOther ||
```
PC98, line 57: We have to reverse all our conditions since now the variable tells us if the step is complete

components/donation/form/DonationPageFormNavigation.tsx
```diff
@@ -42,11 +42,6 @@ const DonationPageFormNavigation: React.FC<Props> = ({
 }) => {
   const { container, icon, positiveMargin } = useStyles();
   const dispatch = useContext(DonationPageStateDispatch);
-  const [maxCurStepIndex, setMaxCurStepIndex] = useState(-1);
```
PC98, line 45: Earlier, we were letting users use the form step bar at the top to navigate to any step that they had previously visited by just clicking on the step bar titles. However, this was leading to a bug. I can explain more in detail on Slack, but just trust me, it's much easier to remove this feature than modify our code to deal with the bug.

components/donation/form/DonationPageFormNavigation.tsx
```diff
@@ -58,7 +53,7 @@ const DonationPageFormNavigation: React.FC<Props> = ({
         onClick={() => {
           dispatch && dispatch(setStep(index));
         }}
-        disabled={index > maxCurStepIndex}
+        disabled={index > curStepIndex}
```
PC98, line 56: We will still let users use the form step bar at the top to move back to a step that occurs before the step they are currently on; the bug doesn't affect this behavior.

package.json
```diff
@@ -40,6 +40,7 @@
     "@types/js-cookie": "^2.2.6",
     "@types/jsonwebtoken": "^8.3.8",
     "@types/material-ui": "^0.21.7",
+    "@types/mongoose": "^5.7.36",
```
PC98, line 43: Just to provide some type hints for the few `.ts` files in `server/`

pages/api/checkToken.ts
```diff
@@ -9,4 +9,4 @@ export default async (
   req: NextApiRequest,
   res: NextApiResponse
 ): Promise<void> =>
-  await handleRequestWithPayloadResponse(req, res, checkToken, ["token"]);
+  handleRequestWithPayloadResponse(req, res, checkToken, ["token"]);
```
PC98, line 12: removed `await` as it was violating the new lint rule!

pages/api/createDonation.ts
```diff
@@ -3,13 +3,13 @@ import { createDonation } from "server/actions/donation";
 import { handleRequestWithPayloadResponse } from "utils/util";
 
 // @route   POST api/createDonation
-// @desc    Creates a Donation
+// @desc    Logs a Donation in our DB; doesn't process the payment nor interacts with Stripe
```
PC98, line 6: Clarified this, and opened an issue that a new dev can take on to rename this endpoint: #79

server/actions/admin.js
```diff
@@ -27,16 +27,18 @@ const jwtSignAdmin = ({ _id, firstName, lastName, email, nonprofitId }) =>
...
   await Mongo();
 
-  const admin = await Admin.findOne({ email });
-  if (admin) {
-    if (!(await bcrypt.compare(password, admin.password))) {
-      throw new Error(errors.admin.INVALID_LOGIN);
-    }
-
-    return jwtSignAdmin(admin);
+  const admin = await Admin.findOne({ email }).lean();
```
PC98, line 30: Better to use `lean` after any Mongoose `find*` function if we don't want the full-fledged document, and just the object is sufficient.

server/actions/donation.ts
```diff
@@ -17,13 +17,13 @@ export async function createDonation({
   email,
   amount,
   nonprofitId
-}: DonationType): Promise<boolean> {
+}: DonationType): Promise<void> {
```
PC98, line 20: The return type is not doing anything/was not being used

server/actions/donation.ts
```diff
@@ -33,10 +33,11 @@ export async function createDonation({
...
   });
 
-  nonprofit.donations.push(donation);
-  nonprofit.save();
-
-  return true;
+  nonprofit
+    .update({
+      $push: { d: donation }
+    })
```
PC98, line 39: let's not push to the array in memory, and use a Mongoose/Mongo method instead.

server/actions/nonprofit.ts
```diff
@@ -29,25 +29,36 @@ export async function createNonprofit({
   });
 }
 
-export async function getNonprofitNames() {
+export async function getNonprofitNamesWithIds(): Query<NonprofitNameWithId[]> {
```
PC98, line 32: A lot of changes based on Mongoose type hints

utils/errors.ts
```diff
@@ -2,14 +2,11 @@ export default {
...
     INVALID_LOGIN: "Incorrect email or password.",
-    INVALID_ORG: "Please enter a valid nonprofit organization.",
     INVALID_TOKEN: "The current token is expired or invalid.",
     USER_EXISTS: "A user with this email already exists."
   },
-  donation: {
-    INVALID_ORG: "The nonprofit does not currently exist in our database."
-  },
   nonprofit: {
-    ALREADY_EXISTS: "There already exists a nonprofit by this name."
```
PC98, line 13: Removed error strings that mean the same thing

server/actions/admin.js
```diff
@@ -27,16 +27,18 @@ const jwtSignAdmin = ({ _id, firstName, lastName, email, nonprofitId }) =>
...
+  if (!admin) {
+    throw new Error(errors.admin.INVALID_LOGIN);
+  }
+  if (Array.isArray(admin)) {
```
PC98, line 34: For some reason, this might be an array according to Mongoose's type definitions. So deal with that explicitly.

PR #110, opened 2020-09-06
Fix secrets scripts for windows and update README

Arthelon:
Windows command prompt users will have to run `yarn secrets:login` and `yarn secrets:sync` to sync secrets


PC98 (approved):
Thanks for fixing this!

PR #111, opened 2020-09-06
Fix migration script for windows users

migueldlr:
Use `cross-env` to handle setting `NODE_ENV`

Tested in cmd:
image

And wsl:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "next",
     "build": "next build",
-    "dev:db:migrate": "NODE_ENV=development yarn ts-node -O '{\"module\": \"commonjs\"}' -r module-alias/register server/migration.ts",
+    "dev:db:migrate": "cross-env NODE_ENV=development yarn ts-node -O {\\\"module\\\":\\\"commonjs\\\"} -r module-alias/register server/migration.ts",
     "start": "next start",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "next",
     "build": "next build",
-    "dev:db:migrate": "NODE_ENV=development yarn ts-node -O '{\"module\": \"commonjs\"}' -r module-alias/register server/migration.ts",
+    "dev:db:migrate": "cross-env NODE_ENV=development yarn ts-node -O {\\\"module\\\":\\\"commonjs\\\"} -r module-alias/register server/migration.ts",
```
PC98, line 8: Do you need all these extra escape characters? Does `-O '{\"module\": \"commonjs\"}'` not work on Windows?

PC98 (approved):

package.json
migueldlr: It doesn't recognize the quotes otherwise:
image

They mention that JSON needs to be extra-escaped when it's passed into a variable in the [cross-env readme](https://github.com/kentcdodds/cross-env#usage), but I guess it applies to the actual script as well

PC98: Thanks for checking!

PR #112, opened 2020-09-06
Updated README for Windows and packages.json secrets scripts

Cullen101:
README now tells windows users to use yarn secrets:login and yarn secrets:sync in order to fully sync. Also, secrets scripts are updated to newer ones.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,9 +10,9 @@
     "lint": "yarn lint:tsc ; yarn lint:eslint",
     "lint:tsc": "tsc -p tsconfig.json --noEmit",
     "lint:eslint": "eslint '**/*.{tsx,ts,js,jsx}' --fix",
-    "secrets": "yarn secrets:logout && yarn secrets:login 'yarn secrets:sync'",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,9 +10,9 @@
     "lint": "yarn lint:tsc ; yarn lint:eslint",
     "lint:tsc": "tsc -p tsconfig.json --noEmit",
     "lint:eslint": "eslint '**/*.{tsx,ts,js,jsx}' --fix",
-    "secrets": "yarn secrets:logout && yarn secrets:login 'yarn secrets:sync'",
+    "secrets": "npm run secrets:logout && cross-env-shell BW_SESSION=`bw login product@bitsofgood.org --raw` \"npm run secrets:sync\"",
```
KoolCards, line 13: Sorry forgot to mention this earlier. We're using yarn exclusively for the project, so you'll need to change the command from `npm run` to just `yarn`.

KoolCards (changes requested):

PC98:
Actually @Arthelon already opened #110 to address this problem. I will merge in that PR so that you can pull those changes and try again.

Issue #93, opened 2020-08-31
[VMS] Create the root component for `/event` page


PC98:
* Create a component, `components/events/EventsPage.tsx`.
* If #91 has been completed, this component should be rendered in the Next.js page corresponding to the `/events` route.
* For now, have this component simply render the `EventsPageLayout.tsx` component. You can pass `null` or empty `div`s for `EventsPageLayout`'s `childeren` and `leftRailComponent`/`sidebarComponent` props.

PR #114, opened 2020-09-07
Add events page

matthewbrown77:
* Creates basic event page.
* Adds event page route to config.ts.


config.ts
```diff
@@ -21,7 +21,8 @@ export default {
   pages: {
     index: "/",
     login: "/login",
-    donate: (path = "[id]"): string => `/donate/${path}`
+    donate: (path = "[id]"): string => `/donate/${path}`,
+    events: "/events/events"
```
matthewbrown77, line 25: Is having an events subfolder here necessary? From my understanding, all of the other `events` front-end code will be under `components`, so there shouldn't be multiple event-related files under the `pages` directory.

pages/events/events.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import { NextPage, GetServerSideProps } from "next";
+
+const EventsPage: NextPage = () => null;
```
matthewbrown77, line 4: This is where the [root component](https://github.com/gtbitsofgood/gen-soln/issues/93) should go.

matthewbrown77:
Forgot to attach page screenshot.
events_page

config.ts
PC98: So the idea is we will at-least have the following routes:
`/events/` -> to view all events
`/events/[id]` -> to view information about an event with a specific ID.

To support the above routes in Next.js, we will have to create a `pages/events/` folder.

PC98 (approved):
@matthewbrown77 Please rename your file to `pages/events/index.tsx` and make appropriate changes in `config.ts` -- we want to support the `events/` route and not `events/events/` 🙂

Feel free to merge after making this change.

matthewbrown77:
> @matthewbrown77 Please rename your file to `pages/events/index.tsx` and make appropriate changes in `config.ts` -- we want to support the `events/` route and not `events/events/` 
> 
> Feel free to merge after making this change.

@PC98 Addressed in 1f4d668. Merging from previous approval.

PC98:
I will follow up with @Arthelon about this -- even I am not a Vercel team member

Issue #97, opened 2020-08-31
[VMS] Component for event listing card


PC98:
* In this task, you are going to create a component that implements the UI for an event card:![Screen Shot 2020-08-31 at 6.28.42 PM.png](https://images.zenhubusercontent.com/5a64edf34b5806bc2bc39e19/93a8a769-897e-4b6f-97f3-8cdddfd7fcdd)
* See Figma for detailed designs.
* This should be a "dumb" UI component, that is, it will receive all the content (like image, heading, body, meta text) that it has to render from its `props`. It itself won't hold any state.

migueldlr:
@PC98 should the component be absolutely sized or fill its container? There are Figma examples for the component being [246x324](https://www.figma.com/file/HMEsyyl6EGJ4hMaGBNtQNe/Donor-and-Volunteer-Experience-%2F-Fall20?node-id=201%3A1685) and [266x237](https://www.figma.com/file/1VBjx7n3zFQMng4YUNX6iz/Horizon-%2F-Web?node-id=242%3A2045)

PC98:
Thanks for bringing this up. I followed up with Jason and he said that the height should be static, 237px. He said the width can be a value in the 240-260px range depending on the grid layout of the page. Based on the layout component that the "EventsListing" page uses (see "_Production" page design in "Donor and Volunteer Experience", aka User Experience, Figma doc), the width should be 246px. Not sure how we want to handle this dynamic width though, since the Core team hasn't built layout components yet, should we take in width as a prop?

PC98:
Also here are some specs related to max-lines and padding values:
<img width="415" alt="Screen Shot 2020-09-08 at 1 27 49 PM" src="https://user-images.githubusercontent.com/18293058/92510311-a3693600-f1d9-11ea-9777-bf815b3f9dc3.png">

PR #117, opened 2020-09-09
Add large event card component

migueldlr:
#97 
This adds the `EventCardLarge` component which we can use to render an event listing with an image, header text, meta text, and body text. It handles text wrapping properly, and has a fixed height and width.

localhost_3000_donate_64357724

> Why did you add `card` and `cardContainer`? What's with this `box-shadow`? Why not use `border`?

The 1px border was pushing content over 2 pixels leaving an ugly gap between the image and the border (I think it has something to do with subpixels or something). To handle it, I just added a `box-shadow` on the div and wrapped it in another div with 1px padding.

Notes:
- Width is fixed at 250px, this could be parameterized in the future
- We still need to add Visby CF font to be Figma-compliant

### DIFF
--- a/components/events/EventCardLarge.tsx
+++ b/components/events/EventCardLarge.tsx
@@ -1,0 +1,109 @@
+import React from "react";
+import clsx from "clsx";
+
+import makeStyles from "@material-ui/core/styles/makeStyles";
+import Typography from "@material-ui/core/Typography";
+
+const useStyles = makeStyles({
+  card: {
+    width: 248,
+    height: 267,
+    borderRadius: 10,
+    overflow: "hidden",
+    backgroundColor: "#FFFFFF",
+    /* 1px borders don't play nice on Chrome, so we use an equivalent
+     * box-shadow and wrap the div in another div */
+    boxShadow: "0 0 0 1px #F0F0F0"
+  },
+  cardContainer: {
+    width: 250,
+    height: 269,
+    padding: 1
+  },
+  image: {
+    display: "block",
+    height: 128,
+    width: "inherit",
+    objectFit: "cover"
+  },
+  content: {
+    padding: "16px 24px"
+  },
+  header: {
+    color: "#333333",
+    fontFamily: "Visby CF, sans-serif",
+    fontSize: 16,
+    fontWeight: 800,
+    lineHeight: "130%",
+    marginBottom: 6,
+    /* line-clamp isn't supported yet, so the next four rules handle it for us */
+    display: "-webkit-box",
+    WebkitBoxOrient: "vertical",
+    WebkitLineClamp: 2,
+    overflow: "hidden"
+  },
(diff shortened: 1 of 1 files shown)


components/events/EventCardLarge.tsx
```diff
@@ -0,0 +1,109 @@
...
+const useStyles = makeStyles({
+  card: {
+    width: 248,
+    height: 267,
+    borderRadius: 10,
+    overflow: "hidden",
+    backgroundColor: "#FFFFFF",
+    /* 1px borders don't play nice on Chrome, so we use an equivalent
+     * box-shadow and wrap the div in another div */
+    boxShadow: "0 0 0 1px #F0F0F0"
```
PC98, line 16: question: do we anticipate using this color for other shadows, or is this a one-off usage for this component only?

components/events/EventCardLarge.tsx
```diff
@@ -0,0 +1,109 @@
...
+    padding: 1
+  },
+  image: {
+    display: "block",
+    height: 128,
+    width: "inherit",
+    objectFit: "cover"
+  },
+  content: {
+    padding: "16px 24px"
```
PC98, line 30: nit: prefer using `paddingHorizontal: 24` and `paddingVertical: 16` for readability (I still get confused if the first value in `padding: "x y"` is vertical or if it's the second) and because TS can check `number` types better than arbitrary strings for CSS styles.

components/events/EventCardLarge.tsx
```diff
@@ -0,0 +1,109 @@
...
+  },
+  header: {
+    color: "#333333",
+    fontFamily: "Visby CF, sans-serif",
```
PC98, line 34: This is okay as a temporary solution, but yes once the core team implements Horizon Typography component, we will not be defining our own font or text color related styles.

PC98 (approved):
I appreciate the level of detail and thoroughness🤘🏻
Thanks for attaching the screenshot!

components/events/EventCardLarge.tsx
migueldlr: Horizon defines it as `gray-5`, not sure where else it gets used

components/events/EventCardLarge.tsx
migueldlr: `paddingHorizontal` and `paddingVertical` aren't supported 😭

PC98: Ah okay, nvm then 😅

components/events/EventCardLarge.tsx
PC98: Okay, if that's the case we (i.e. Core team) may have to pull it out into a CSS variable or something or add it to our material-ui's theme object. For now, it's fine, we will make this change as part of a larger change when we switch to Horizon components.

PR #120, opened 2020-09-10
Creates EventsPageLayout component

matthewbrown77:
Here is an image of the layout rendered on the `/events` page with scrollbars showing. I used this to verify that each container was independently scrollable.
event_page

The current version is below and has the scrollbars hidden, but the divs are still scrollable (I used the above to test since it's hard to tell when the divs are just a block of the same color).
hidden_scrollbar


PC98 (approved):
Thanks for taking up this task! It will help unblock several other tasks!

PR #121, opened 2020-09-10
event schema created

Fattimo:
event schema created. TODO: users schema for volunteers needed to be integrated


Fattimo:
potentially add another field for "other events like this"

server/models/event.js
```diff
@@ -0,0 +1,51 @@
+const mongoose = require("mongoose");
+const Schema = mongoose.Schema;
+
+// TODO: Volunteer schema
+// Keep in sync with utils/types Donation
```
PC98, line 5: Let's remove these comments, based on our discussion on Slack

server/models/event.js
```diff
@@ -0,0 +1,51 @@
...
+const eventSchema = new Schema({
+  name: {
+    type: String,
+    required: true
+  },
+  nonprofitId: {
+    type: String,
```
PC98, line 12: Can we change this type to also be `Schema.ObjectId`?

server/models/event.js
```diff
@@ -0,0 +1,51 @@
...
+    ref: "Nonprofit",
+    required: true
+  },
+  startDate: {
+    type: Date,
+    required: true,
+    default: Date.now
```
PC98, line 19: Question: does it make sense to have a `default` value for a `required` attribute? Same question applies for `endDate`.

server/models/event.js
```diff
@@ -0,0 +1,51 @@
...
+  },
+  endDate: {
+    type: Date,
+    required: true,
+    default: Date.now
+  },
+  about: {
```
PC98, line 26: I would say let's either mark this as `required`, or default to the empty string. Otherwise on the front-end/while querying for the data, we will have to check for the existence of `about` data.

server/models/event.js
```diff
@@ -0,0 +1,51 @@
...
+    type: Number,
+    required: true
+  },
+  volunteers: [
+    {
+      type: mongoose.Schema.ObjectId,
+      ref: "Volunteer" //TODO
+    }
+  ],
+  image: {
```
PC98, line 39: Let's mark this as required. Based on Figma designs, it seems each event will have an `image`.

server/models/event.js
```diff
@@ -0,0 +1,51 @@
...
+  about: {
+    type: String
+  },
+  maxVolunteers: {
+    type: Number,
+    required: true
+  },
+  volunteers: [
+    {
+      type: mongoose.Schema.ObjectId,
```
PC98, line 35: nit: this can just be `Schema.ObjectId` since we have `const Schema = mongoose.Schema` at the top

server/models/event.js
```diff
@@ -0,0 +1,51 @@
...
+      ref: "Volunteer" //TODO
```
PC98, line 36: Based on Slack discussions, might have to update this.

PC98 (changes requested):
Looks good overall! Left some minor feedback :)

server/models/event.js
Fattimo: Yup! Just as a warning, the schema for donations (model/donation.js) followed the format I used here, so let me know if it would be cool to change the same thing on donation.js and I can add that in as part of this PR.

server/models/event.js
Fattimo: I actually think that a default image would be better. From a user experience, I think setting images for planned events is really annoying, and something I think users may do is create a bunch of events to set a schedule, not really beef them up with images or long descriptions, and then go back when the date is nearer to edit make those volunteer opportunities a lot more flashy. Let me know what type of placeholder image I should find if you end up agreeing with me!

server/models/event.js
PC98: Oh okay. Let's avoid doing that as part of this PR, and if you want you can open a ZenHub issue for this, and attach the "DMS", "backend" labels. Thanks for pointing this out!

server/models/event.js
PC98: Yeah I think that makes a lot of sense! We don't have an official default image right now, but maybe you can use the takeoff thumbnail from Figma.

PC98:
Question: What are some pros/cons of making `startDate` and `endDate` default to `Date.now` as opposed to just `required`?

Fattimo:
> Question: What are some pros/cons of making `startDate` and `endDate` default to `Date.now` as opposed to just `required`?

My original line of thinking was for the virtual to work, we need a start and end date, but after thinking about it a little more, making it required would still force the virtual to work. So, there's really not that big of a difference. 

The one difference I would pose would be 2 less required things for the user to set, in case they just need a placeholder date. If we want users to have a bare minimum of a general date going into the creation of their event, then I can see required being better.

PC98 (approved):
Oh okay, makes sense. But maybe we should default `endDate` to be like `Date.now` + 1 hour? Or we can change this later if we notice that users are frequently creating 1-hour duration events.

PR #122, opened 2020-09-10
Creates root component for event page

matthewbrown77:


pages/events/index.tsx
```diff
@@ -1,7 +1,10 @@
 import React from "react";
 import { NextPage, GetServerSideProps } from "next";
+import EventsPage from "components/events/EventsPage";
 
-const EventsPage: NextPage = () => null;
+const EventsNextPage: NextPage = () => {
```
matthewbrown77, line 5: Renamed to avoid conflict with imported `EventsPage`.

components/events/EventsPage.tsx
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import EventsPageLayout from "components/events/EventsPageLayout";
+
+const EventsPage: React.FC = () => {
+  return <EventsPageLayout sidebarComponent={null}>{null}</EventsPageLayout>;
```
PC98, line 5: nit: you might be able to get away with just `<EventsPageLayout sidebarComponent={null} />` 👀

PC98 (approved):
🚀

Issue #89, opened 2020-08-31
[VMS] Create a migration script to seed Events


PC98:
* After the Event schema is fully fleshed out, let's add some dummy events to our local dev DB.
* We will do this through a migration script so that other developers can also run the script to populate their dev DBs. First, ensure that your DB already has all existing migrations applied:
```bash
yarn dev:db:migrate status
```
* If any of the migrations are "PENDING", you can run them all in a single command: `yarn dev:db:migrate up`
* Now, create a new migration:
```bash
yarn dev:db:migrate seed-events
```
* This should have created a new file, `server/migrations/somenumber-seed-events.ts`
* You can delete the initial contents of the file and follow the code in `server/migrations/20200829032827-seed-nonprofits.ts` to insert Events in the `db`.
* Make sure you populate all required fields, make any necessary changes to the "nonprofits" collection so that the many-to-many relations work out. Be sure to also populate the address field, use Google Maps to find the latitude and longitude coordinates of addresses you add. 
* **In this script, you can let every events' `volunteers` array be empty.** Once the User schema is ready, we will write another migration script that populates the `volunteers` array. 
* Test your migration by running `yarn dev:db:migrate up`.

PR #125, opened 2020-09-10
Add getUpcomingEvents action

matthewbrown77:
Testing this query locally properly returned the 5 mock `Event` objects with the closest dates.
query_results

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -1,0 +1,23 @@
+import Mongo from "server/index";
+import Event from "server/models/event";
+
+export async function getUpcomingEvents(): Promise<Array<Event>> {
+  await Mongo();
+
+  const result = await Event.aggregate([
+    {
+      $project: {
+        difference: {
+          $abs: {
+            $subtract: [new Date(), "$startDate"]
+          }
+        },
+        doc: "$$ROOT"
+      }
+    },
+    { $sort: { difference: 1 } },
+    { $limit: 5 }
+  ]);
+
+  return result;
+}
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -1,23 +1,18 @@
 import Mongo from "server/index";
 import Event from "server/models/event";
+import { Event as EventType } from "utils/types";
 
-export async function getUpcomingEvents(): Promise<Array<Event>> {
+export async function getUpcomingEvents(): Promise<EventType[]> {
   await Mongo();
 
-  const result = await Event.aggregate([
-    {
-      $project: {
-        difference: {
-          $abs: {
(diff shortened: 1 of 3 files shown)


server/actions/events.ts
```diff
@@ -0,0 +1,23 @@
...
+          $abs: {
+            $subtract: [new Date(), "$startDate"]
+          }
+        },
+        doc: "$$ROOT"
+      }
+    },
+    { $sort: { difference: 1 } },
+    { $limit: 5 }
+  ]);
```
PC98, line 20: I think using `aggregate` is a bit overkill here. Can we use `find` along with `$gte`? Example of working with dates: https://mongoosejs.com/docs/tutorials/dates.html#querying 

We should also consider using the [`lean()` operation](https://mongoosejs.com/docs/tutorials/lean.html). However, the return type `Promise<Array<Event>>` won't make sense then. You can either omit the return type OR create an `Event` interface in `utils/types.ts` corresponding to the `Event` Mongoose model (see the `Nonprofit` interface as an example), import the interface in this file as `EventType`, and use `Promise<EventType[]>` as the return type.

PC98 (commented):
If you wrote queries to create this dummy event data in your local DB, it might be helpful to write a migration script so that other people can use the same data! (this is totally optional though but will help whoever takes on #89)

server/actions/events.ts
matthewbrown77: I didn't think I could accomplish the query using only `find()`, but that actually isn't true. Swapped over to using `find()` and added `Event` type in d076c96.

utils/types.ts
```diff
@@ -28,6 +29,17 @@ export interface Donation {
   timestamp: Date;
 }
 
+export interface Event {
```
PC98, line 32: nit: consider adding a comment above this line, `// Keep in sync with the backend schema` and a similar comment in `server/models/event.js`: `// Keep in sync with utils/types Event`.

utils/types.ts
```diff
@@ -28,6 +29,17 @@ export interface Donation {
...
+  name: string;
+  nonprofitId: string;
+  startDate: Date;
+  endDate: Date;
+  about: string;
+  maxVolunteers: number;
+  volunteers: Array<mongoose.Types.ObjectId>;
```
PC98, line 39: as an FYI, this is fine for now but when we decide to send `Event` data to the front-end, we will probably cast `ObjectId` to `string` so that the data can be sent via JSON.

PC98 (approved):
Thanks for completing another task! 🔥

matthewbrown77:
Comments addressed in 01d9fdc. Merging from previous approval.

Issue #86, opened 2020-08-31
[VMS] Modify User and Nonprofit schema to establish relationships with Event schema


PC98:
* In #85, we added the `nonprofitId` and `volunteers` fields to the Event schema, but we also have to:

- [x] Make changes to the User schema to model the many-to-many relationship.
- [x] Make changes to the Nonprofit schema to model the one-to-many relationship.

* Please see links in #85 describing many-to-many and one-to-many relationships for implementation details.

PR #129, opened 2020-09-13
#86 Modify User and Nonprofit schema to establish relationships with Event schema

Fattimo:
added event relationship to nonprofit.js

### DIFF
--- a/server/models/nonprofit.js
+++ b/server/models/nonprofit.js
@@ -47,8 +47,13 @@ const nonprofitSchema = new Schema({
     ref: "Donation"
   },
   events: {
-    type: [ObjectId],
-    ref: "Events"
+    type: [
+      {
+        type: Schema.ObjectId,
+        ref: "Events"
+      }
+    ],
+    default: []
   }
 });
 
--- a/server/models/nonprofit.js
+++ b/server/models/nonprofit.js
@@ -45,6 +45,10 @@ const nonprofitSchema = new Schema({
   donations: {
     type: [ObjectId],
     ref: "Donation"
+  },
+  events: {
+    type: [ObjectId],
+    ref: "Events"
   }
 });
 
--- a/server/migrations/20200913210455-migrate-nonprofits-add-events.ts
+++ b/server/migrations/20200913210455-migrate-nonprofits-add-events.ts
@@ -1,0 +1,14 @@
+import { Db } from "mongodb";
+import { MigrationFunction } from "migrate-mongo";
+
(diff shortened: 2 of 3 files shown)


Fattimo:
Right now, the User.js still doesn't exist. How should I handle that part of the issue? 

Also, I don't have the clearest of idea of how we're going to be using nonprofit.js. Would it make sense to default donations or events to the empty array?

PC98:
* Yes, let's have `events` default to the empty array. I was planning on opening a PR to make `donations` default to an empty array as well. Since these changes involve changes to a schema (Nonprofit schema) for which we have data in our DB, you should write an accompanying migration script. You can base your migration script off of mine when I open the PR.
* Yeah, I guess we are blocked on Users. I will add the "blocked" label to this issue and move the second part to Sprint 2.

PC98:
@Fattimo have a look at #130 for how to write a migration script. Your script will be very similar, except:
* You will use `events` instead of `donations`
* Your `update` clauses in the `up` and `down` functions will query all data, that is, `db.collection("nonprofits").update({}, /* other stuff*/)`. My migration script was querying for only certain entries since my code change modified an existing field to have a default value, but your migration script has to update all documents to contain a new field.
* Use a different name for your script, `yarn dev:db:migrate create migrate-nonprofits-add-events`, or whatever name you think is appropriate.

Feel free to follow me on Slack if this migrations jargon is not making sense!

PC98:
@Fattimo As an FYI, I merged #130 and resolved conflicts in your branch. Before you work on adding the migration script, please pull the latest changes from your branch, `git pull origin matt/86-user_np_to_event`

server/models/nonprofit.js
```diff
@@ -47,8 +47,13 @@ const nonprofitSchema = new Schema({
     ref: "Donation"
   },
   events: {
-    type: [ObjectId],
-    ref: "Events"
+    type: [
+      {
+        type: Schema.ObjectId,
+        ref: "Events"
```
Fattimo: Typo in ref: "Events", should be "Event", but I caught it in the merge resolution commit, so the change is there.

Fattimo:
> @Fattimo As an FYI, I merged #130 and resolved conflicts in your branch. Before you work on adding the migration script, please pull the latest changes from your branch, `git pull origin matt/86-user_np_to_event`

I saw this just now, but I think I worked it out after the fact. My merge commit additionally updates a small typo.

PC98 (approved):

PR #131, opened 2020-09-13
Add migration script for events

matthewbrown77:
Migration result.
events_migration


server/migrations/20200913143420-seed-events.ts
```diff
@@ -0,0 +1,69 @@
...
+        coordinates: [35.09096, -85.18751]
+      }
+    },
+    maxVolunteers: 10,
+    volunteers: []
+  }
+];
+
+export const up: MigrationFunction = async (db: Db) => {
+  await db.collection("events").insertMany(events);
```
PC98, line 62: In the `up` and `down` functions, can we also modify the nonprofit document's `events` array to include these events? In #129, Matt added the `events` field to the Nonprofit schema (I realize #129 was merged after you opened this PR, but it would be great if you could do this 🙂)

PC98 (commented):
Thanks for doing this! It will be really cumbersome to seed more events in our DB, I think for future seed scripts we should use: https://github.com/marak/Faker.js/ I will make Zenhub tasks to add and use this library

server/migrations/20200913143420-seed-events.ts
matthewbrown77: To keep things simple for adding the `event` documents to the relevant `events` field, I defined the `_id` myself instead of letting mongo create it.

PC98: Ah okay sure. One minor thing: we are defaulting Event IDs to be of type `ObjectID` and not String. Can you check if this works:
```javascript
import {ObjectID} from "mongodb";
// your code
  _id: new ObjectID("81409375"),
//
```

server/migrations/20200913143420-seed-events.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  for (const event of events) {
+    await db
+      .collection("nonprofits")
+      .updateOne({ _id: event.nonprofitId }, { $set: { events: event._id } });
```
PC98, line 67: Can you make sure the `events` field is set to an array `event._id`, and not set to the `event._id` itself? Alternatively, consider using `push`: https://docs.mongodb.com/manual/reference/operator/update/push/

server/migrations/20200913143420-seed-events.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+};
+
+export const down: MigrationFunction = async (db: Db) => {
+  await db
+    .collection("events")
+    .deleteMany({ _id: { $in: events.map(_ => _._id) } });
+
+  await db.collection("nonprofits").updateMany({}, { $set: { events: [] } });
```
PC98, line 76: Ideally, here you should follow an approach similar to `up`, where you loop through `events` and only remove those event IDs that you added. But if this is too complicated, then never mind.

matthewbrown77:
@PC98 The changes in f14d93f properly add and remove `events` in the nonprofit documents using `$push` and `$pull`. I tested these changes running the `up` migrate command on a fresh db instance.

PC98 (approved):

PR #132, opened 2020-09-13
added stripeAccount field to nonprofit collection

Cullen101:
added stripeAccount field on types.ts, nonprofit.js, and nonprofit.ts


KoolCards (approved):

PC98 (changes requested):
Let's hold off on merging this, I will add some feedback soon

utils/types.ts
```diff
@@ -17,6 +17,7 @@ export interface Nonprofit {
   logo: string;
   primaryColor: string;
   secondaryColor: string;
+  stripeAccount: string;
```
PC98, line 20: Question: will the `stripeAccount` field be confidential?

Should we be sending this data to our front-end? If not, can you remove this attribute from this type (the type here represents all the fields the front-end can access) and modify the `getNonprofitById()` server action to exclude this field when returning data to the front-end?

PC98 (commented):
Can you also write a migration script? In general, every schema change for models that already have data in our DB should be accompanied by a migration script. You will have to:
* Create a new migration script, `yarn dev:db:migrate migrate-nonprofits-add-stripe-account` or whatever name you think is appropriate.
* Have a look at existing scripts in `server/migrations` to implement the migration `up` and migration `down` functions. In `up`, you will want to update each document in the `"nonprofits"` collection to have a `stripeAccount` field; for now, you can use dummy values or an empty string for `stripeAccount`. In `down` you will do the opposite - remove this field from all documents.

utils/types.ts
KoolCards: The workflow for creating a new stripe account is that when a nonprofit onboards, they'll need to send their stripeAccount id from the frontend to the backend. However, I think you're right we actually probably shouldn't be sending this back to the fronted.

KoolCards:
> Can you also write a migration script? In general, every schema change for models that already have data in our DB should be accompanied by a migration script. You will have to:
> 
> * Create a new migration script, `yarn dev:db:migrate migrate-nonprofits-add-stripe-account` or whatever name you think is appropriate.
> * Have a look at existing scripts in `server/migrations` to implement the migration `up` and migration `down` functions. In `up`, you will want to update each document in the `"nonprofits"` collection to have a `stripeAccount` field; for now, you can use dummy values or an empty string for `stripeAccount`. In `down` you will do the opposite - remove this field from all documents.

Ahh ok. So I had assumed that since we weren't updating the stripeAccount field of existing nonprofit collections that we wouldn't need to create a migration script. However, I guess ur saying that it's best to have dummy values as the stripeAccount for now.

utils/types.ts
PC98: Thanks for confirming -- we should still remove it from the type as well since the front-end uses this type to assume the shape of the returned data from the back-end. I will open a PR to update the comments in this file to make this clear.

server/actions/nonprofit.ts
```diff
@@ -45,8 +47,9 @@ export async function getNonprofitById(_id: string): Promise<NonprofitType> {
   await Mongo();
 
   // Exclude donation information for now:
-  const nonprofit = await Nonprofit.findOne({ _id }, { donations: 0 }).lean();
-
+  const nonprofit = await Nonprofit.findOne({ _id }, { donations: 0 })
```
KoolCards, line 50: Last thing, you can specify the field to exclude as the projection parameter in the mongoose query. Can you modify this so that we exclude stripeAccount the same way we exclude donations?

server/migrations/20200913194527-migrate-nonprofits-add-stripe-account.ts
```diff
@@ -0,0 +1,17 @@
+import { Db } from "mongodb";
+import { MigrationFunction } from "migrate-mongo";
+
+const nonprofits = ["46099993", "64357724", "92232372"];
+export const up: MigrationFunction = async (db: Db) => {
+  await db
+    .collection("nonprofits")
+    .updateMany({ _id: { $in: nonprofits } }, { $set: { stripeAccount: "" } });
```
PC98, line 8: Just do:
```
db
    .collection("nonprofits")
    .update(
      { },
      { $set: { stripeAccount: "" } },
      { multi: true }
    );
```
This will let you avoid hard-coding the non-profit IDs. You can also take this approach in your `down()` function

utils/types.ts
PC98: @Cullen101 I think you forgot to delete this

PC98 (commented):
You will also have a conflict with the develop branch -- let me know if you need help resolving that!

server/migrations/20200913194527-migrate-nonprofits-add-stripe-account.ts
KoolCards: I thought we'd only want to update only the current 3 nonprofits with the migration script because the schema change will automatically update any future nonprofits we add to the db. That's why it's hardcoded, but I guess either way works.

PC98: Yeah, ideally we shouldn't hardcode. Right now I am thinking that all the "seed" migration scripts won't run on prod DB, but all the "migrate" migration scripts will. Since this script will run on prod DB, here is what we can to ensure that if a document already has a `stripeAccount`, we don't set it to the empty string:
```
await db
  .collection("nonprofits")
  .update(
    { stripeAccount: { $exists: false } },
    { $set: { stripeAccount: [] } },
    { multi: true }
  );
```
And similar changes to `down`.

We will *most* of the data to our prod DB using server actions triggered by API calls.

KoolCards:
@PC98 re-requesting your review before merging. We made stripeAccount an optional field in the Nonprofit type, as stripeAccount has to be sent from frontend to backend when onboarding in the Nonprofit object, but never from backend to frontend.

PC98 (approved):
Tbh you can still omit this from the `Nonprofit` interface imo, since we might not end up using the interface in the form when collecting data and seding it to the backend (though we won't know for sure until the form code is actually implemented).

I just don't any future front-end devs confused and think that they can access the `stripeAccount` from a Nonprofit type even though that's not true/the DB never returns it.

server/actions/nonprofit.ts
```diff
@@ -15,7 +15,8 @@ export async function createNonprofit({
   background,
   logo,
   primaryColor,
-  secondaryColor
+  secondaryColor,
+  stripeAccount
 }: NonprofitType) {
```
PC98, line 20: If line 19 was throwing a TS error when you removed `stripeAccount`, you can just use `NextApiRequest["body"]` instead of `NonprofitType`; I don't think annotating with `NonprofitType` was giving us anything nice.

Issue #103, opened 2020-09-01
[DMS] Add createStripeAccount nonprofit action


KoolCards:
Currently, there isn't a way to create a Stripe account for nonprofits who use our website. We need to create a function that does this, so that nonprofits can receive payments. We need to create a connected Stripe account for the nonprofit by redirecting the nonprofit to the Stripe web form (basically we'd redirect the user to a site like [this](https://dashboard.stripe.com/oauth/authorize?response_type=code&client_id=ca_GzSc2LsvquOYz0PviCx3AiSsMorLoN6B&scope=read_write)). 

- [x] Create a new nonprofit action (server/actions/nonprofit.ts) called `createStripeAccount`. You'll likely have to also create a new Stripe object in the nonprofit.ts file (look at server/actions/nonprofit.ts). You'll have to parse the account object that stripe returns for the account number and return this value. [Step 2.1](https://stripe.com/docs/connect/enable-payment-acceptance-guide)
- [x] Create a new nonprofit action (server/actions/nonprofit.ts) called `linkStripeAccount`. This function should accept a Stripe account number as the parameter. [Step 2.2](https://stripe.com/docs/connect/enable-payment-acceptance-guide)

The `linkStripeAccount` function itself should do 2 things.
1. Create an account link using the **connected** account id as the paramter. For now, just have the refresh and redirect url's point to back to the donate page in `config.ts`. 
2. Parse whatever Stripe returns from the account link for the url to redirect the user to.

- [x] Test your functions by calling `createStripeAccount` to get an account number and then calling `linkStripeAccount` for a url. Go to this url in a private browser and ensure that it looks similar to [this](https://dashboard.stripe.com/oauth/authorize?response_type=code&client_id=ca_GzSc2LsvquOYz0PviCx3AiSsMorLoN6B&scope=read_write).

ML-Chen:
For the first subtask, here's what I've written so far. I'm not sure if this is what it is asking for, especially the part about "create a new Standard account with the bank account information".
```
import Stripe from "stripe";

/* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
const stripe = new Stripe(config.stripe.secret_key!, {
  apiVersion: "2020-03-02"
});

// ...

export async function createStripeAccount() {
  const account = await stripe.accounts.create({
    type: "standard"
  });

  return account.id;
}
```

KoolCards:
[...]

PR #138, opened 2020-09-13
Add actions createStripeAccount() and linkStripeAccount()

ML-Chen:
#103

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -4,3 +4,4 @@
 DEV_DB_NAME=test
 DEV_DB_URL=mongodb://localhost
 NEXT_PUBLIC_DEV_STRIPE_PUBLISHABLE=pk_test_gM41BNgzCQfEG0G5CgPsPx7v00345tjQxr
+DEV_BASE_URL=http://localhost:3000
--- a/config.ts
+++ b/config.ts
@@ -20,9 +20,7 @@ export default {
   jwtSecret: isDevEnv
     ? process.env.DEV_JWT_SECRET
     : process.env.PROD_JWT_SECRET,
-  baseUrl:
-    (isDevEnv ? process.env.DEV_BASE_URL : process.env.PROD_BASE_URL) ||
-    "http://localhost:3000",
+  baseUrl: isDevEnv ? process.env.DEV_BASE_URL : process.env.PROD_BASE_URL,
   pages: {
     index: "/",
     login: "/login",
--- a/config.ts
+++ b/config.ts
@@ -20,6 +20,9 @@ export default {
   jwtSecret: isDevEnv
     ? process.env.DEV_JWT_SECRET
     : process.env.PROD_JWT_SECRET,
+  baseUrl:
+    (isDevEnv ? process.env.DEV_BASE_URL : process.env.PROD_BASE_URL) ||
+    "http://localhost:3000",
   pages: {
     index: "/",
     login: "/login",
--- a/server/actions/nonprofit.ts
+++ b/server/actions/nonprofit.ts
@@ -95,9 +95,13 @@ export async function linkStripeAccount(
   const accountLink = await stripe.accountLinks.create({
     account: accountId,
     // Placeholder URLs
-    refresh_url: `http://localhost:3000/${config.pages.donate(defaultNonprofitId)}`, // prettier-ignore
-    return_url: `http://localhost:3000/${config.pages.donate(defaultNonprofitId)}`, // prettier-ignore
+    refresh_url: `http://localhost:3000${config.pages.donate(defaultNonprofitId)}`, // prettier-ignore
+    return_url: `http://localhost:3000${config.pages.donate(defaultNonprofitId)}`, // prettier-ignore
     type: "account_onboarding"
   });
   return accountLink.url;
 }
+
+createStripeAccount()
+  .then(accountId => linkStripeAccount(accountId).then(url => console.log(url)))
+  .catch(err => console.log(err));
--- a/server/actions/nonprofit.ts
+++ b/server/actions/nonprofit.ts
@@ -94,14 +94,9 @@ export async function linkStripeAccount(
   const defaultNonprofitId = await getDefaultNonprofitId();
   const accountLink = await stripe.accountLinks.create({
     account: accountId,
-    // Placeholder URLs
     refresh_url: config.baseUrl + config.pages.donate(defaultNonprofitId),
     return_url: config.baseUrl + config.pages.donate(defaultNonprofitId),
     type: "account_onboarding"
   });
   return accountLink.url;
 }
-
-createStripeAccount()
-  .then(accountId => linkStripeAccount(accountId).then(url => console.log(url)))
-  .catch(err => console.log(err));
--- a/server/actions/nonprofit.ts
+++ b/server/actions/nonprofit.ts
@@ -1,11 +1,18 @@
 import Mongo from "server/index";
+import Stripe from "stripe";
 import Nonprofit from "server/models/nonprofit";
 import errors from "utils/errors";
 import { Nonprofit as NonprofitType } from "utils/types";
 import { Query } from "mongoose";
+import config from "../../config";
 
 type NonprofitNameWithId = Pick<NonprofitType, "name"> &
   Pick<NonprofitType, "_id">;
+
+/* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+const stripe = new Stripe(config.stripe.secret_key!, {
+  apiVersion: "2020-03-02"
+});
 
 export async function createNonprofit({
   name,
@@ -72,4 +79,25 @@ export async function getDefaultNonprofitId(): Promise<string> {
   }
 
   return result[0]._id;
+}
+
+export async function createStripeAccount(): Promise<Stripe.Account["id"]> {
+  const account = await stripe.accounts.create({
+    type: "standard"
+  });
+  return account.id;
+}
+
+export async function linkStripeAccount(
+  accountId: Stripe.Account["id"]
+): Promise<Stripe.AccountLink["url"]> {
+  const defaultNonprofitId = await getDefaultNonprofitId();
+  const accountLink = await stripe.accountLinks.create({
+    account: accountId,
+    // Placeholder URLs
+    refresh_url: `http://localhost:3000/${config.pages.donate(defaultNonprofitId)}`, // prettier-ignore
+    return_url: `http://localhost:3000/${config.pages.donate(defaultNonprofitId)}`, // prettier-ignore
+    type: "account_onboarding"
+  });
+  return accountLink.url;
 }
--- a/server/actions/nonprofit.ts
+++ b/server/actions/nonprofit.ts
@@ -95,8 +95,8 @@ export async function linkStripeAccount(
   const accountLink = await stripe.accountLinks.create({
     account: accountId,
     // Placeholder URLs
-    refresh_url: `http://localhost:3000${config.pages.donate(defaultNonprofitId)}`, // prettier-ignore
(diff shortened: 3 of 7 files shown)


server/actions/nonprofit.ts
```diff
@@ -1,12 +1,19 @@
 import Mongo from "server/index";
+import Stripe from "stripe";
 import Nonprofit from "server/models/nonprofit";
 import errors from "utils/errors";
 import { Nonprofit as NonprofitType } from "utils/types";
 import { Query } from "mongoose";
+import config from "../../config";
```
PC98, line 7: nit: `import config from "config";`

server/actions/nonprofit.ts
```diff
@@ -1,12 +1,19 @@
...
 
 type NonprofitNameWithId = Pick<NonprofitType, "name"> &
   Pick<NonprofitType, "_id">;
 
+/* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+const stripe = new Stripe(config.stripe.secret_key!, {
+  apiVersion: "2020-03-02"
+});
```
PC98, line 15: Can you put this code in a function, perhaps in `server/index.js`? I want to avoid code duplication since `donation.ts` also creates a Stripe object with this `apiVersion`

server/actions/nonprofit.ts
```diff
@@ -73,3 +80,28 @@ export async function getDefaultNonprofitId(): Promise<string> {
...
+    refresh_url: config.baseUrl + config.pages.donate(defaultNonprofitId),
+    return_url: config.baseUrl + config.pages.donate(defaultNonprofitId),
+    type: "account_onboarding"
+  });
+  return accountLink.url;
+}
+
+createStripeAccount()
+  .then(accountId => linkStripeAccount(accountId).then(url => console.log(url)))
+  .catch(err => console.log(err));
```
PC98, line 107: should this be here? I think you meant to use these lines for testing purposes only.

server/actions/nonprofit.ts
```diff
@@ -73,3 +80,28 @@ export async function getDefaultNonprofitId(): Promise<string> {
...
+  return account.id;
+}
+
+export async function linkStripeAccount(
+  accountId: Stripe.Account["id"]
+): Promise<Stripe.AccountLink["url"]> {
+  const defaultNonprofitId = await getDefaultNonprofitId();
+  const accountLink = await stripe.accountLinks.create({
+    account: accountId,
+    // Placeholder URLs
```
PC98, line 97: Can you put a `TODO` in your comment, and potentially create a task to track this.

Also just so I understand things correctly, what's the URL that we actually want to use for `refresh_url` and `return_url`?

config.ts
```diff
@@ -20,6 +20,9 @@ export default {
   jwtSecret: isDevEnv
     ? process.env.DEV_JWT_SECRET
     : process.env.PROD_JWT_SECRET,
+  baseUrl:
+    (isDevEnv ? process.env.DEV_BASE_URL : process.env.PROD_BASE_URL) ||
+    "http://localhost:3000",
```
PC98, line 25: I think the `|| "http://localhost:3000"` is redundant (assuming that our URLs are not empty or null/undefined)

.env.development
```diff
@@ -4,3 +4,4 @@
 DEV_DB_NAME=test
 DEV_DB_URL=mongodb://localhost
 NEXT_PUBLIC_DEV_STRIPE_PUBLISHABLE=pk_test_gM41BNgzCQfEG0G5CgPsPx7v00345tjQxr
+DEV_BASE_URL=http://localhost:3000
```
PC98, line 7: What do you think of adding `PROD_BASE_URL` as well, since it's not really a "secret". You could either add `PROD_BASE_URL` in a new `.env.production` file, or to simplify things, you can rename this file to `.env` and add `PROD_BASE_URL` here as well.

PC98 (changes requested):

config.ts
ML-Chen: That's true, but I didn't want to have to write `/* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */` each time we use it.

server/actions/nonprofit.ts
ML-Chen: I've just pushed a commit that removes this part.

server/actions/nonprofit.ts
ML-Chen: A function to construct a Stripe object with that apiVersion? Or do we want a stripe singleton?

server/actions/nonprofit.ts
ML-Chen: I'm not sure what the URL is supposed to be. @KoolCards ?

.env.development
ML-Chen: What is the PROD_BASE_URL supposed to be? Should I just write `PROD_BASE_URL=` with no value?

server/actions/nonprofit.ts
ML-Chen: Fixed

server/actions/nonprofit.ts
KoolCards: We don't have an actual `refresh_url` and `return_url` yet because these are supposed to be the urls to our nonprofit onboarding form (which doesn't exist right now). Right now, we just have it redirect to the donate page.

.env.development
PC98: Won't `PROD_BASE_URL` just be where we have our web app deployed? `https://bog-gen-soln.vercel.app`?
So yeah, you can either add `PROD_BASE_URL=https://bog-gen-soln.vercel.app` to a `.env.production` file, or to keep things simpler rename this file to `.env` and add it here (I prefer the latter).

config.ts
PC98: Tbh I don't imagine that we will use `baseUrl` that much; our codebase hadn't needed it so far 🙂 
Also as an FYI, you can also use `/* eslint-disable @typescript-eslint/no-non-null-assertion */` to completely disable the lint rule in a block of text and enable it after the block: `/* eslint-enable @typescript-eslint/no-non-null-assertion */` 

I just don't like hardcoding `http://localhost:3000` twice since if we ever change the default port, it's possible someone updates the .env file but forgets to update `config.ts`.

server/actions/nonprofit.ts
PC98: Is there a way to easily implement a singleton in JS? I am not that familiar, but if you know a way that is not hacky then sure go for it. Otherwise, a simple function that creates and returns a Stripe object should do.

server/actions/nonprofit.ts
ML-Chen: Added a comment so now it looks like this:
```ts
  const accountLink = await stripe.accountLinks.create({
    account: accountId,
    /* TODO: These are placeholder URLs. They should be replaced with the URL of
         our nonprofit on-boarding form once we create that. */
    refresh_url: config.baseUrl + config.pages.donate(defaultNonprofitId),
    return_url: config.baseUrl + config.pages.donate(defaultNonprofitId),
    type: "account_onboarding"
  });
```

PC98: Ah I see. But we are not re-directing to that particular non-profit's donate page here, just to the "default" non-profit's? I would recommend either re-directing to the particular non-profit's donation page or to the index page (given by the base URL).

I don't want to introduce further usage of `getDefaultNonprofitId()`; this function will go away when our index page has been built.

PC98: @ML-Chen Thanks for adding the TODO -- bonus points if you want to expand your TODO to mention that the URLs eventually need to be replaced with the nonprofit onboarding form URL

server/actions/nonprofit.ts
ML-Chen: By singleton I meant instantiating the Stripe object once in index.js and importing it into other files. I think I was using the word "singleton" wrong.

PC98: No I think I understood what you meant. I think the problem is that if you do this in `index.js`:
```
export const stripe = new Stripe(config.stripe.secret_key, {
  apiVersion: "2020-03-02"
});
```
this code will run whenever someone imports anything from `index.js`, even if they are not importing the `stripe` object (which is basically all files in `server/actions`)

server/actions/nonprofit.ts
ML-Chen: Okay I'll change it to this then:
```ts
  const accountLink = await stripe.accountLinks.create({
    account: accountId,
    /* TODO: These are placeholder URLs. They should be replaced with the URL of
         our nonprofit on-boarding form once we create that. */
    refresh_url: config.baseUrl,
    return_url: config.baseUrl,
    type: "account_onboarding"
  });
```

config.ts
ML-Chen: What about this?
```ts
  baseUrl: (isDevEnv
    ? process.env.DEV_BASE_URL
    : process.env.PROD_BASE_URL) as string,
```

.env.development
ML-Chen: So would that include merging .env.development.local and .env.local?

config.ts
PC98: This will further hide the error? Since I think `null as string` will just cast the string to `"null"`. Whereas with `/* eslint-disable @typescript-eslint/no-non-null-assertion */` we can track where the potential null casting is taking place.

.env.development
PC98: I think `.env.development.local` and `.env.local` can remain unchanged.

server/actions/nonprofit.ts
ML-Chen: So https://stackoverflow.com/questions/37325667/does-es6-module-importing-execute-the-code-inside-the-imported-file an imported file is only supposed to be run once even if it is imported multiple times. When I have a console.log statement in nonprofit.ts, it runs twice the first time I load a page after build, and then once for every subsequent page load. Right now for testing purposes I have a function from nonprofit.ts imported into three files.

.env.development
ML-Chen: So then `.env` is for both production and development environment variables that are not secret, and `.env.local` is for production secrets? That seems a little confusing but I guess it's alright.

PC98: Yeah, we need at least two env files. One for secrets and one for non-secrets.

We can combine `.env.local` and `.env.development.local` but this will be more involved -- the yarn secrets scripts will have to be modified and we will have to ask all devs to delete their local copy of `.env.development.local`

server/actions/nonprofit.ts
PC98: > Right now for testing purposes I have a function from nonprofit.ts imported into three files.

Not sure what you mean.
If it is being run multiple times, I assume you will be taking the approach of creating a function that constructs the Stripe object?

config.ts
ML-Chen: What about
```ts
baseUrl: (isDevEnv ? process.env.DEV_BASE_URL : process.env.PROD_BASE_URL)!
```

Although now that I'm not concatenating the baseUrl anymore, TypeScript doesn't care that it's `string | undefined`

PC98: Sure, `baseUrl: (isDevEnv ? process.env.DEV_BASE_URL : process.env.PROD_BASE_URL)!` is fine. Or if TS doesn't care, you can just skip this and be consistent with how other variables in `config` are defined.

server/actions/nonprofit.ts
ML-Chen: It seems that when I have a console.log statement in index.js, it prints every time I go to a different page, so it seems that yeah it does get run an awful lot of times.

.env.development
ML-Chen: I've made changes that should resolve this and your other comments. Please see the new commit.

PC98 (approved):
Thanks for iterating on this!

Issue #71, opened 2020-08-28
[CORE] Check if the entered password recovery email exists in DB


PC98:
* We have a Recover Password page where users can enter their email.
* After a user enters their email and clicks "Get Password", we should check if the entered email exists in our database or not.
* If it exists,  we will let the app behave as it already does. However, if the email does not exist, we should inform the user about the error. This task will involve both front-end and back-end code.
#### Backend
- [x] Create a new error string in `utils/errors.ts` for the `admin` object. Call it `INVALID_EMAIL_FOR_PASSWORD_RECOVERY`. You can assign it the value `"This email is not associated with a user account."`
- [x] In `server/actions/admin.js`, create a new function and call it `recoverPassword` (just use whatever name you think is appropriate).
- [x] This function signature will look a lot like that of the already implemented `login` function, except that it will only take `{ email }` as input.
- [x] The function should check if an `Admin` with the given `email` exists in our DB (hint: use `Admin.exists({ email })`).
- [x] If `true`, return `true`, otherwise throw an `Error` with the `INVALID_EMAIL_FOR_PASSWORD_RECOVERY` message. As an FYI, The code you write in this function will run on the server-side, and thus you are able to query our DB directly.
- [x] You will **NOT** implement the actual logic of sending a password recovery email, but please put a TODO comment (`// TODO`) explaining that a password recovery email should be sent to the provided `email`.
- [x] Create a new route in `pages/api/`, call it something like `recoverPassword.ts` 
- [x] Look at how `pages/api/login.ts` is implemented. You will have to do something very similar in the `recoverPassword.ts` file; you will call `handleRequestWithPayloadResponse` but with `recoverPassword` (the method you created in `server/actions/admin.js`) as your first argument and `["email"]` as the second argument.
- [x] The front-end code will make an HTTP POST request to this endpoint. Be sure to add your endpoint under the `apis` object in `config.ts`.
#### Frontend
- [x] In `requests/admin.ts`, write a function, `recoverPassword`. This will look a lot like `login`, except you will call `fetchRequestWithPayloadResponse` with `urls.apis.recoverPassword` as the first argument and your `body` will be `JSON.stringify({ email })`.
- [x] Look at `onPressCTA` in `LoginFormContent.tsx`. Notice how it makes a request to `login`, catches the error, and calls `setError` to display the error message to the user. You will have to do something very similar in `RecoverPasswordFormContent.tsx`:
```javascript
case "INITIAL":
  try {
[...]

PR #139, opened 2020-09-14
Added functionality for checking if recovery email exists

awang378:
#71 
Added backend and frontend logic to check if recovery password email exists in the database and handle accordingly.


components/auth/form/RecoverPasswordFormContent.tsx
```diff
@@ -22,8 +25,9 @@ const RecoverPasswordFormContent: React.FC<ContentComponentProps> = ({
   navigateToContent
 }) => {
   // TODO: Use this to show an error message below the email input field
-  const [hasError] = useState(false);
+  //const [hasError] = useState(false);
```
PC98, line 28: Nit: delete these two comments.

components/auth/form/RecoverPasswordFormContent.tsx
```diff
@@ -35,14 +39,25 @@ const RecoverPasswordFormContent: React.FC<ContentComponentProps> = ({
...
+            if (
+              err instanceof Error &&
+              err.message === errors.admin.INVALID_EMAIL_FOR_PASSWORD_RECOVERY
+            ) {
+              message = errors.admin.INVALID_EMAIL_FOR_PASSWORD_RECOVERY;
+            }
+            setError(message ?? errors.GENERIC_TEXT);
+          } finally {
+            stopLoading();
+          }
```
PC98, line 59: Curios: when you tested out the UI, did React throw any error in the browser console? I wasn't sure if calling `stopLoading()` in the finally after `setCurrentState` is called would work or not.

server/actions/admin.js
```diff
@@ -43,6 +43,17 @@ export async function login({ email, password }) {
...
 
+export async function recoverPassword({ email }) {
+  await Mongo();
+  const admin = await Admin.exists({ email });
+  if (admin) {
+    //TODO password recovery email sent to provided email
+    return true;
+  } else {
+    throw new Error(errors.admin.INVALID_EMAIL_FOR_PASSWORD_RECOVERY);
+  }
```
PC98, line 54: nit: can throw directly, no need for the `else` block.

requests/admin.ts
```diff
@@ -19,3 +19,12 @@ export const checkTokenRequest = async (token: string) =>
     },
     body: JSON.stringify({ token })
   });
+
+export const recoverPassword = async (email: string): Promise<string> =>
+  fetchRequestWithPayloadResponse<string>(urls.apis.recoverPassword, {
```
PC98, line 24: Should be type `fetchRequestWithPayloadResponse<boolean>` since your server action returns `Promise<boolean>`

server/actions/admin.js
```diff
@@ -43,6 +43,17 @@ export async function login({ email, password }) {
   return jwtSignAdmin(admin);
 }
 
+export async function recoverPassword({ email }) {
+  await Mongo();
+  const admin = await Admin.exists({ email });
+  if (admin) {
```
PC98, line 49: nit: technically this is not an admin object but just a boolean that indicates if an Admin with that email exists or not. You should either rename this variable to something like `doesAdminExist`, or you can in-line it: `if (await Admin.exists({ email }))`

components/auth/form/RecoverPasswordFormContent.tsx
awang378: React didn't throw any error in the console when I was testing.

PC98 (approved):

PR #140, opened 2020-09-14
Matt/98 return near events

Fattimo:
added a script in sever/actions/events.ts to return 5 nearest events to a specified coordinate.

typo in branch name, should be matt/98


server/models/event.js
```diff
@@ -44,7 +44,7 @@ const eventSchema = new Schema({
     },
     location: {
       type: pointSchema,
-      index: "2dshere",
+      index: "2dsphere",
```
PC98, line 47: good catch!

server/actions/events.ts
```diff
@@ -16,3 +16,26 @@ export async function getUpcomingEvents(): Promise<EventType[]> {
...
+  await Mongo();
+
+  const result = (await Event.find({})
+    .where("address.location")
+    .near({
+      center: [long, lat],
+      sptherical: true
+    })
+    .lean()
+    .limit(5)) as EventType[];
```
PC98, line 38: were you able to test this? If not it's fine and we can wait until #131 is merged. That will allow you to seed your local DB with dummy data and test this query.

server/models/nonprofit.js
```diff
@@ -45,6 +45,10 @@ const nonprofitSchema = new Schema({
   donations: {
     type: [ObjectId],
     ref: "Donation"
+  },
+  events: {
+    type: [ObjectId],
+    ref: "Events"
```
PC98, line 51: Can you pull `develop` into your branch? This indicates that you probably don't have the latest changes.

server/actions/events.ts
Fattimo: Not yet, so once #131 is merged I will get onto testing!

server/actions/events.ts
```diff
@@ -16,3 +16,26 @@ export async function getUpcomingEvents(): Promise<EventType[]> {
...
+  const result = (await Event.find({})
+    .where("address.location")
+    .near({
+      center: [long, lat],
+      spherical: true
+    })
+    .lean()
+    .limit(5)) as EventType[];
+
+  return result;
```
PC98, line 40: For some reason I didn't notice this before, but let's try to avoid `return await`: https://eslint.org/docs/rules/no-return-await

The code in `getNearestEvents()` and `getUpcomingEvents()` use unncecessary `return await` (the assignment to `result` variable just hides this). 
Here is what I would recommend:
```
export async function getUpcomingEvents() { // we can skip the type annotation here
  await Mongo();

  const result = Event.find({
    startDate: {
      $gte: new Date()
    }
  })
    .lean()
    .sort({ startDate: 1 })
    .limit(5)
    .exec(); // exec will convert the Mongoose Query to a fully-fledged JavaScript Promise

  return  result as Promise<EventType[]>; // casting 👀
}
```

Fattimo:
image
image from console logs resulting from testing. distances/lats&longs checked against an online source

PC98 (approved):
Thanks for iterating on this!!

PR #147, opened 2020-09-15
Fix Windows compatibility issues with `yarn lint`

migueldlr:
This change fixes `yarn lint` on Windows

cmd:
Screenshot 2020-09-15 193239

wsl:
Screenshot 2020-09-15 193443


package.json
```diff
@@ -9,9 +9,9 @@
     "dev:db:migrate": "yarn run:ts-node:script server/migration.ts",
     "run:server:action": "yarn run:ts-node:script server/run-action.ts",
     "start": "next start",
-    "lint": "yarn lint:tsc ; yarn lint:eslint",
-    "lint:tsc": "tsc -p tsconfig.json --noEmit",
-    "lint:eslint": "eslint '**/*.{tsx,ts,js,jsx}' --fix",
+    "lint": "yarn lint:tsc && yarn lint:eslint",
+    "lint:tsc": "tsc --noEmit",
```
PC98, line 13: Question: do we need _all_ these changes to make the script work on Windows or is escaping the quotes enough?

PC98 (approved):

package.json
migueldlr: After some investigation, it seems we only need to change lines 12 and 14. [TS docs](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) suggest that we could remove the `-p` option on line 13 without affecting functionality, but we don't have to if you're against it

PC98: This is fine, thanks for checking 🙂

PR #148, opened 2020-09-16
modular orange button

savannahjoyner:
orange button component

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "dev:db:migrate": "cross-env NODE_ENV=development yarn ts-node -O {\\\"module\\\":\\\"commonjs\\\"} -r module-alias/register server/migration.ts",
     "start": "next start",
     "lint": "yarn lint:tsc ; yarn lint:eslint",
-    "lint:tsc": "tsc -p tsconfig.json --noEmit",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "dev:db:migrate": "cross-env NODE_ENV=development yarn ts-node -O {\\\"module\\\":\\\"commonjs\\\"} -r module-alias/register server/migration.ts",
     "start": "next start",
     "lint": "yarn lint:tsc ; yarn lint:eslint",
-    "lint:tsc": "tsc -p tsconfig.json --noEmit",
+    "lint:tsc": "tsc --noEmit",
```
PC98, line 11: You have merge conflicts in this file, let me know if you need help resolving them!

PC98 (commented):
Thanks for working on this 🙂
A few comments:
* Could you rename this component to `HorizonButton.tsx`? We are trying to establish a convention where all Horizon library components begin with the work Horizon.
* Could you maybe attach screenshots showing this button? I know this might be hard since we don't have a way to test components in isolation right now. I would recommend going through the codebase and replacing one of the button components (like the one used in `AuthPageForm.tsx` with this Button to illustrate what it looks like.

PR #151, opened 2020-09-17
Changed href of login link on donation form

parthnagpal17:
Changed the href on the login links in the donation form @ML-Chen @KoolCards

### DIFF
--- a/components/donation/form/DonationPageFormAdminLoginLink.tsx
+++ b/components/donation/form/DonationPageFormAdminLoginLink.tsx
@@ -6,7 +6,7 @@ import Typography from "@material-ui/core/Typography";
 
 import HorizonLink from "@horizon/HorizonLink";
 
-import config from "../../../config";
+import config from "config";
 
 const useStyles = makeStyles({
   container: {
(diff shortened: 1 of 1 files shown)


components/donation/form/DonationPageFormAdminLoginLink.tsx
```diff
@@ -6,6 +6,8 @@ import Typography from "@material-ui/core/Typography";
 
 import HorizonLink from "@horizon/HorizonLink";
 
+import Config from "../../../config";
```
PC98, line 9: just do `import config from "config"` please, we have set-up module aliasing 🙂

parthnagpal17: Just fixed it!

PC98: I think you forgot to fix the path 😅
The idea is that using relative paths like `from "../../../config";` is not a good idea/makes code harder to maintain and refactor.
That's why, in our repo, we let you do `from "config"` whenever importing the config object.

parthnagpal17: Oh, my bad. I got it now, thanks :)

PC98 (approved):

PR #161, opened 2020-09-19
Typography

aniruddhamurali:
I implemented the typography from the Horizon Web / Design Figma (added the Visby CF and Open Sans fonts).


utils/theme.ts
```diff
@@ -46,6 +47,36 @@ const createNonprofitMuiTheme = (nonprofit: Nonprofit | undefined): Theme =>
...
+      },
+      body2: {
+        fontFamily: "Open Sans",
+        fontSize: "18px"
+      },
+      caption: {
+        fontFamily: "Open Sans",
+        fontSize: "16px"
+      }
+    },
```
PC98, line 79: While this works, I think the expectation is to create a new component in `components/core/`, called `HorizonTypography`. Under the hood, `HorizonTypography` will render material-ui's `Typography` component and all call sites in our code will use `HorizonTypography` instead of directly using material-ui's `Typography`.

Using `HorizonTypography` props types, you can restrict what all variants you want to support, like "h1", "h2", "caption", etc. Then the file `HorizonTypography.tsx` itself will contain all these styles for each variant.

Have a look at this code: https://gist.github.com/PC98/9949493ca38f26f9f4aa76c240c5b0c8 for inspiration This code uses React Native, but I just want it to give you an idea of how to do this (though keep in mind you won't be using `StyleSheet`, that's a React Native thing, you will be using `makeStyles` and possibly `clsx`, see `componets/ButtonWithLowercaseText` and `components/ButtonWithNonprofitColor`).

PC98 (commented):
thanks for working on this! left some comments

rithikgavvala (approved):

PC98:
@aniruddhamurali Now that I have thought about this more, I think your approach of modifying material-ui's theme object does make more sense. But we still want to create a different component for  "HorizonTypography". What do you think of this approach? https://gist.github.com/PC98/edf6c0f2e78525eaca4a4869a834507a The `typographyStyles` object can be imported in `utils/theme.ts`

Also, we should perhaps put our font files in the `public/` folder; this is what Next.js recommends for static file serving. We can also move `fonts.css` to `public/` and then import it using `<link href="/fonts.css" rel="stylesheet" />` in `_document.tsx` (or `_app.tsx` if that doesn't work).

aniruddhamurali:
@PC98 Sorry about the delay! I appreciate the awesome feedback. I think your approach works really well. I haven't had any experience with Material UI before so I looked at a couple tutorials and the material ui documentation to figure out how to do the typography stuff. I agree that we should move `fonts.css` and the font files to the `public/`. I'll get working on that!

aniruddhamurali:
@PC98 I tried my best to incorporate the feedback you gave me and pushed a new commit. I created the HorizonTypography component in the `components/core/` folder and I moved the font files to `public/`. I have a couple of questions:
- In `HorizonTypography.tsx`, I was wondering what @horizon was? It says that it could not find the horizon module. I assume you wanted to import a certain shade of gray from the module, but since I don't know about the module, I simply put the color `"gray"` in commonStyles for now.
- I was curious to learn about what the Props and HorizonTypography variables in `HorizonTypography.tsx` were for. I tried to add them but it I get this error: 
```
Type 'FC<Props>' is not assignable to type 'TypographyOptions | ((palette: Palette) => TypographyOptions) | undefined'.
  Type 'FunctionComponent<Props>' is not assignable to type '(palette: Palette) => TypographyOptions'.
    Types of parameters 'props' and 'palette' are incompatible.
      Type 'Palette' is not assignable to type 'PropsWithChildren<Props>'.
        Property 'variant' is missing in type 'Palette' but required in type '{ variant: Variant; }'
```

PC98 (approved):

PC98:
Yep, I will add the colors later, so that's fine.
`@horizon` is simply an alias to the `components/core` path, see this: https://github.com/GTBitsOfGood/gen-soln/blob/develop/tsconfig.json#L22 Not sure why this wasn't working for you :/

Hm I will have look at the types issue later. 
`HorizonTypography` is simply a component we will create and use instead of directly using material-ui's `Typography` component.

PR #163, opened 2020-09-20
Add getEventById action

matthewbrown77:
Result when passing a valid id:
get_event_by_id

Result when id is not found:
id_not_found

One issue I ran into when testing this PR was with the event `_id` field. Since this field was previously set as strings in the migration script, the `findById` call could never correctly find an event since this method implicitly casts the string input as an `ObjectId` and thus the comparison between the string and objects were never equal. to fix this, I changed the migration script to set the `_id` field by creating a new `ObjectId` (which is why the id passed into the method and the actual id on the event object returned are different). However, I don't think is a good long-term solution since the id actually stored in the database looks very different. One option is to assign the `_id` field in the schema directly as a string (similar to the nonprofit schema). 

Thoughts? @PC98

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -1,5 +1,6 @@
 import Mongo from "server/index";
 import Event from "server/models/event";
+import errors from "utils/errors";
 import { Event as EventType } from "utils/types";
 
 export async function getUpcomingEvents() {
@@ -37,4 +38,16 @@ export async function getNearestEvents({ lat, long }: Coordinates) {
     .exec();
 
   return result as Promise<EventType[]>;
+}
+
+export async function getEventById(_id: string): Promise<EventType> {
+  await Mongo();
+
+  const event = (await Event.findById(_id).lean()) as EventType;
+
+  if (event == null) {
+    throw new Error(errors.event.INVALID_ID);
+  }
+
+  return event;
 }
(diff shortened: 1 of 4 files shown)


PC98:
> One option is to assign the _id field in the schema directly as a string (similar to the nonprofit schema).

Sure, let's do that. This will let you avoid modifying your migration script and make the Event object easier to work with. We should generally never modify an existing migration script.

I thin you can modify add to the Event schema:
```javascript
  _id: {
    type: String,
    default: () => new mongoose.Types.ObjectId().toHexString()
  },
```
The `default` part will ensure that when creating an `Event` object (through a Mongoose query) we won't have to specify the ID ourselves. 

If you make this change, you will have to modify the Nonprofit schema to use `type: Schema.ObjectId` where it specifies `ref: "Event"`.

matthewbrown77:
@PC98 Added the _id field to the nonprofit schema in aa2bac4. The nonprofit schema already specified type: Schema.ObjectId, so no change necessary there. Confirmed the query worked as intended with these changes.

PC98:
> The nonprofit schema already specified type: Schema.ObjectId
I think we will have to change it to `String` no? Since our Event IDs are strings now.

matthewbrown77:
Changed in 5af5c61.

PC98 (approved):
💯

PR #167, opened 2020-09-20
Add getAllEventIds action

matthewbrown77:
Action output:
get_all_event_ids

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -55,7 +55,5 @@ export async function getEventById(_id: string): Promise<EventType> {
 export async function getAllEventIds(): Promise<string[]> {
   await Mongo();
 
-  const result = (await Event.find({}, { _id: 1 })).map(event => event._id);
-
-  return result;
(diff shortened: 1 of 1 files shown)


server/actions/events.ts
```diff
@@ -51,3 +51,11 @@ export async function getEventById(_id: string): Promise<EventType> {
 
   return event;
 }
+
+export async function getAllEventIds(): Promise<string[]> {
+  await Mongo();
+
+  const result = (await Event.find({}, { _id: 1 })).map(event => event._id);
+
+  return result;
```
PC98, line 60: to keep things consistent and concise, consider using `return Event.distinct("_id").exec();` (this might also help speed things up since we don't have to load all Event objects in memory).

matthewbrown77: Good call. I didn't see that in the nonprofit actions file, but that is much cleaner.

PC98 (approved):

PR #175, opened 2020-09-21
adding auth0 to the project

rithikgavvala:
added auth0


package.json
```diff
@@ -51,6 +52,7 @@
     "@typescript-eslint/eslint-plugin": "^3.10.1",
     "@typescript-eslint/parser": "^3.10.1",
     "babel-eslint": "^10.1.0",
+    "commander": "^6.1.0",
```
PC98, line 55: Seems you have conflicts with the develop branch since `commander` is already a dependency

PC98 (commented):
@rithikgavvala If we intend on following this: https://github.com/GTBitsOfGood/nextjs-sso-starter/
It seems the only auth related dependency we will need is `next-auth`.

Will also have to do `yarn add -D @types/next-auth` to get types of TS.

PC98:
@rithikgavvala Like I said, you might want to remove `@auth0/auth0-react` as a dependency since based on: https://github.com/GTBitsOfGood/nextjs-sso-starter/, we might only need `next-auth` as a dependency. You also have conflicts in `yarn.lock`. Here is what I would recommend, while you are in your branch:
```
yarn remove @auth0/auth0-react
yarn add next-auth
git pull origin develop
yarn install
```
And then commit+push your changes

PC98 (approved):

PR #177, opened 2020-09-21
Add time filter UI

### LINKED ISSUE #134: [VMS] Build UI for the Time filter
* Create a component corresponding to the Time filter, shown below the Location filter in the [Figma designs](https://www.figma.com/file/HMEsyyl6EGJ4hMaGBNtQNe/Donor-and-Volunteer-Experience-Fall20?node-id=0%3A1).
* You can hard code the various options, **skip the "Choose Dates" option** for now.
* Just like the location filter, we will want to update the URL based on the filter options selected by the user:
  * Go to https://careers.google.com/jobs/results/ and select checkboxes under "Job types" to understand I am talking about.
  * Have a look at the code for the Location filter for inspiration (if it has been implemented) but we basically want to use [`useRouter`](https://nextjs.org/docs/api-reference/next/router).
  * If possible, we will only use the URL to store, read and write to the filter state as opposed to using `useState`. See the description of #96 for a hook that can be used to do this.
  * As a suggestion, the options for this filter can be represented as type `Dropdown[]` (import this from `utils/types`):
```javascript
const timeOptions: Dropdown[] = [
  {text: "Today", value: "TODAY"},
  {text: "Tomorrow", value: "TOMORROW"},
  // so on and so forth
 ];
```
  * Above, `text` will be used when options are rendered in the sidebar UI while `value` will be used to update our URL query parameters.
* You will also have to implement functionality to minimize the Time filter (see the down arrow icon in the Figma design):
  * Get the SVG corresponding to the icon from the [Horizon Figma doc](https://www.figma.com/file/1VBjx7n3zFQMng4YUNX6iz/Horizon-Web?node-id=318%3A529).
  * Use the [`SvgIcon`](https://material-ui.com/components/icons/#svgicon) component.
 * Helpful links to look at for checkbox component and minimization functionality: [Checkbox](https://material-ui.com/components/checkboxes/), [minimization](https://material-ui.com/api/collapse/#collapse-api) and possibly: https://material-ui.com/components/lists/#nested-list and https://material-ui.com/components/lists/#checkbox.
* **Bonus points if you implement a component that can be easily extended to provide the checkbox and minimization functionality**. This component can then be used by both the Time filter and the Cause filter.

migueldlr:
This change adds the time filter UI and some rudimentary functionality with routing (closes #134). To do this, we make a couple of changes:

- Refactor content of `EventsPageLeftRailComponent.tsx` to include a header and several `EventsPageLeftRailFilter`s. Each `EventsPageLeftRailFilter` will correspond to its own filter - Location, Time, Cause, etc.
- Create `EventsPageLeftRailFilter.tsx`, which is given a header and content
  - Make these collapsible - we use a `collapsed` state to do this, and handle accessibility using a new helper component `FocusVisibleOnly.tsx`
- Create `EventsPageTimeFilter.tsx`, which gets passed as the content of an `EventsPageLeftRailFilter`
  - These checkboxes use `useRouter` to modify the URL of the page
  - To control the checkbox UI, we use `CheckedIcon` and `UncheckedIcon`, which are based off of Figma SVG exports

The issue mentions making the time filter component reusable, and in the future we can take the checkbox logic/appearance content out of `EventsPageTimeFilter` and move it into a new generic component.

Screencap here:
https://streamable.com/8on414


components/events/EventsPageTimeFilter.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const router = useRouter();
+  const {
+    root,
+    option,
+    optionRoot,
+    optionLabel,
+    optionLabelSelected
+  } = useStyles();
+
+  const times: string[] =
```
PC98, line 70: I think you can just let TS infer `string[]` for you

components/events/EventsPageTimeFilter.tsx
```diff
@@ -0,0 +1,118 @@
...
+    router.query.time == null
+      ? []
+      : !Array.isArray(router.query.time)
+      ? [router.query.time]
+      : router.query.time;
+
+  const select = async (selected: string) => {
+    await router.push({
+      query: times.includes(selected)
+        ? { time: times.filter((s: string) => s !== selected) }
```
PC98, line 80: same here, TS should infer the type of `s` for you.

components/events/EventsPageTimeFilter.tsx
```diff
@@ -0,0 +1,118 @@
...
+        : { time: [...times, selected] }
+    });
+  };
+
+  return (
+    <div className={root}>
+      <FormGroup className={optionRoot}>
+        {timeOptions.map(timeOption => (
```
PC98, line 88: nit: can destructure `timeOption` here

components/events/EventsPageTimeFilter.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <FormControlLabel
+            key={timeOption.text}
+            label={
+              <Typography
+                className={clsx(
+                  optionLabel,
+                  times.includes(timeOption.value) && optionLabelSelected
```
PC98, line 95: Might make sense to declare a variable in this scope, `const isOptionChecked = times.includes(timeOption.value)` and then use it since it seems you are using this value more than once in the scope.

components/events/EventsPageTimeFilter.tsx
```diff
@@ -0,0 +1,118 @@
...
+              >
+                {timeOption.text}
+              </Typography>
+            }
+            className={option}
+            control={
+              <Checkbox
+                className={optionLabel}
+                checked={times.includes(timeOption.value)}
+                onChange={() => select(timeOption.value)}
```
PC98, line 106: As a suggestion, Checkbox's `onChange` function receives an event object that can be used to pull out the new checked state, `event.target.checked`. You can pass this information to `select` and avoid `times.includes(selected)` in there.

components/events/EventsPageTimeFilter.tsx
```diff
@@ -0,0 +1,118 @@
...
+      : !Array.isArray(router.query.time)
+      ? [router.query.time]
+      : router.query.time;
+
+  const select = async (selected: string) => {
+    await router.push({
+      query: times.includes(selected)
+        ? { time: times.filter((s: string) => s !== selected) }
+        : { time: [...times, selected] }
+    });
```
PC98, line 82: I imagine we will be re-using code like this for other filters as well. This PR is already long, so no need to refactor now. Maybe I should create a task to create some sort of common hook that provides this functionality of reading writing to `router.query`. Thoughts?

PC98 (commented):
Left some minor comments but this is really awesome and impressive!!!

components/events/EventsPageTimeFilter.tsx
PC98, line 70: Doesn't TS infer this type for you?

components/events/EventsPageTimeFilter.tsx
migueldlr: Didn't want to do this because it means I have to write out `return` but you're probably right haha

components/events/EventsPageTimeFilter.tsx
migueldlr: I think I'd pull the whole component out including the content and styling as well. We could have the logic in a hook, but I don't know if it's necessary because we only have one other similar component.

components/events/EventsPageTimeFilter.tsx
migueldlr: I think I'm gonna keep the `includes()` check in `select` - we want to derive all state from the URL params rather than any DOM state, and our source of truth here should be the URL.

components/events/EventsPageTimeFilter.tsx
PC98: Yeah, we could do both. The idea with the hook is that it will abstract away details of `router.query` from our UI/component code. The hook can just expose simple utilities to read, put and remove from `router.query` given a key. The hook itself will also be used by location filter, which is different than the checbox filter abstraction we create.

I will create a task for this so that someone can work on it

PC98 (approved):
🔥

PR #179, opened 2020-09-22
Fix config import syntax

### LINKED ISSUE #72: [CORE] Use consistent syntax when importing `config` object
* Different files in the codebase import the `config` object differently. For instance, `LoginFormContent.tsx` imports it as:
```javascript
import urls from "config";
```
* Let's be consistent, and always import `config` object as:
```javascript
import config from "config";
```

ayush-goyal:
Fixes config import syntax and keeps it consistent as
```javascript
import config from "config";
```

Closes #72


PC98 (approved):
💯

PR #180, opened 2020-09-22
Create stripe object with nonprofit stripe account

KoolCards:
- Creates the stripe object using the nonprofit's stripe account.
- @Cullen101 changes should be merged soon after to ensure that the donation functionality still works.


pages/donate/[id].tsx
```diff
@@ -33,6 +46,9 @@ export const getStaticProps: GetStaticProps = async ({ params }) => {
     getNonprofitById(id)
   ]);
 
+  const stripeAccount = nonprofit.stripeAccount;
+  delete nonprofit.stripeAccount;
```
Comment (PC98) [inline pages/donate/[id].tsx:50]: If it is safe to send `stripeAccount` information to the client, I don't think you need this? Like you can just let it be part of the `nonprofit` object and on line 27, read the information from there. If we do make this change, I think we can make `stripeAccount` a non-optional field in `interface Nonprofit`

pages/donate/[id].tsx
```diff
@@ -13,9 +13,22 @@ import DonationPage from "components/donation/DonationPage";
...
+
+interface Props {
+  stripeAccount: string;
+}
+
+const NonprofitDonationPage: NextPage<
+  Props & React.ComponentProps<typeof DonationPage>
+> = ({ stripeAccount, ...props }) => (
+  <Elements
+    stripe={loadStripe(config.stripe.publishable_key!, { stripeAccount })}
```
Comment (PC98) [inline pages/donate/[id].tsx:27]: nit: maybe add a `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion ` here

pages/donate/[id].tsx
Comment (KoolCards) [reply pages/donate/[id].tsx]: agreed it's probably easier just to have the stripeAccount field info all in the client.

pages/_app.tsx
```diff
@@ -5,13 +5,8 @@ import { ThemeProvider } from "@material-ui/core/styles";
 import CssBaseline from "@material-ui/core/CssBaseline";
 import createNonprofitMuiTheme from "utils/theme";
 import { Nonprofit } from "utils/types";
-import { Elements } from "@stripe/react-stripe-js";
-import { loadStripe } from "@stripe/stripe-js";
 import config from "config";
```
PC98, line 8: Nit: delete the config import line

components/donation/DonationPage.tsx
```diff
@@ -34,6 +34,7 @@ const DonationPage: React.FC<Props & DropdownProps> = ({
         <DonationPageMainContent
           description={nonprofit.about}
           selectedNonprofitId={dropdownProps.selectedValue}
+          stripeAccount={nonprofit.stripeAccount}
```
PC98, line 37: Any reason we need `stripeAccount` in this component, `DonationPageMainContent` and `DonationPageForm`? For this PR at least I think we don't need it?

pages/donate/[id].tsx
```diff
@@ -13,9 +13,25 @@ import DonationPage from "components/donation/DonationPage";
...
+import { loadStripe } from "@stripe/stripe-js";
+
+interface Props {
+  stripeAccount: string;
+}
+
+const NonprofitDonationPage: NextPage<
+  Props & React.ComponentProps<typeof DonationPage>
+> = props => (
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
Comment (PC98) [inline pages/donate/[id].tsx:26]: Ah I think the comment doesn't work since it is on line 26  and the lint warning is on line 28. You either do `// eslint-disable @typescript-eslint/no-non-null-assertion` on line 26 and then `// eslint-enable @typescript-eslint/no-non-null-assertion` on line 34, OR try to put your existing comment immediately before line 28. But yeah this is not a big deal

pages/donate/[id].tsx
```diff
@@ -13,9 +13,25 @@ import DonationPage from "components/donation/DonationPage";
...
 
-const NonprofitDonationPage: NextPage<React.ComponentProps<
-  typeof DonationPage
->> = props => <DonationPage {...props} />;
+import { Elements } from "@stripe/react-stripe-js";
+import { loadStripe } from "@stripe/stripe-js";
+
+interface Props {
+  stripeAccount: string;
+}
```
Comment (PC98) [inline pages/donate/[id].tsx:21]: You are not actually returning `stripeAccount` in `getStaticPaths`. TS isn't complaining since `NonprofitDonationPage` is not typed correctlty, but #184 should fix it.

After #184 goes in, you can remove this `interface Props` and access `stripeAccount` via `props.nonprofit.stripeAccount` inside `NonprofitDonationPage`

components/donation/DonationPage.tsx
KoolCards: I'm adding this (even though its unused right now) because the PR that Cullen is working on will use this new field.

pages/donate/[id].tsx
Comment (KoolCards) [reply pages/donate/[id].tsx:26]: the lint pre-commit hook made the Elements tag multiple lines :(. good catch tho

pages/donate/[id].tsx
Comment (KoolCards) [reply pages/donate/[id].tsx]: Ohh, forgot to remove this. This was for the previous commit, where I created a separate props field for the stripeAccount.

PC98 (approved):
💯

Issue #83, opened 2020-08-30
[DMS] Regenerate donate page if the nonprofit id doesn't exist


KoolCards:
Currently, the donate page (`pages/donate/[id].tsx`) uses `getStaticPaths` to retrieve a list of nonprofit names at build time (when the website is deployed), and `getStaticProps` to pre-render donate pages for the nonprofits at build time. However, this means we need to manually re-build the website each time a new nonprofit onboards onto our website to generate the new donate page. Instead, we should attempt to rebuild the donate page for any nonprofit id and throw an error if the nonprofit id doesn't exist.

- [x] Look at the Next.js [docs on data fetching](https://nextjs.org/docs/basic-features/data-fetching) to understand how `getStaticPaths()` and `getStaticProps()` work together. Read about the [`fallback` property](https://nextjs.org/docs/basic-features/data-fetching#fallback-true) in detail - you should modify how it is being used in `pages/donate/[id].tsx` to complete this task.
- [x] Notice that the `getStaticProps()` method calls `getNonprofitById(id)` to retrieve the non-profit object from our DB. When `fallback` was set to `false`, we could be certain that a non-profit with  id=`id` existed in our DB. But now, Next.js won't render a 404 page by default for non-existant `id`s. So you must enclose this code in a `try-catch` to catch [run-time exceptions thrown by `getNonprofitById(id)`](https://github.com/GTBitsOfGood/gen-soln/blob/develop/server/actions/nonprofit.ts#L51) for non-existant `id`s.
- [x] In your `try` block, you can keep returning `props: { nonprofit, items, selectedValue: id }`, however, you will have to decide what you want to return in the `catch` block. You could return an empty object, `props: {}`, as a suggestion.
- [x] Now, we will have to deal with type changes. Since `NonprofitDonationPage` is no longer guaranteed to receive a `nonprofit` object from `getStaticProps()` every time, it's **incorrect** to say:
```javascript
const NonprofitDonationPage: NextPage<React.ComponentProps<typeof DonationPage >>
```
So first, let's modify our `getStaticProps` function signature to:
```javascript
// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
export const getStaticProps = async (context: GetStaticPropsContext) => {
```
And we will modify the `NonprofitDonationPage` declaration to:
```javascript
const NonprofitDonationPage: NextPage<InferGetStaticPropsType<typeof getStaticProps>> = props => {
```
[...]

PR #183, opened 2020-09-22
Add page for the /event/[id] route

### LINKED ISSUE #159: [VMS] Add a Next.js page for the `/events/[id]` route
* Just like we have the `donate/[id].tsx` component, we will have to make a component corresponding to the `/event/[id]` route.
* The UI can just render `null`.
* Be sure to implement the `getStaticPaths` methods which will call the server action that returns all Event IDs.
* Be sure to implement `getStaticProps` which will call a server action to get information about a specific event and return the Event object.
* **Your code will look very similar to `donate/[id].tsx`, however `donate/[id].tsx` is undergoing changes as part of #83**. These are good changes, and we want to incorporate these changes for the `/events/[id]` route as well. If #83 has been closed by the time you start working on this issue, then you can copy code. Otherwise, message me and we will figure out how else to incorporate those changes.
* You should also modify the `pages` object in `config.ts`. We can either change `pages.events` to be a function, like `pages.donate`, or we can add a new field, `pages.event`:
```javascript
donate: (path = "[id]"): string => `/donate/${path}`,
events: "/events",
event: (path = "[id]"): string => `/events/${path}`,
```
We should talk about the pros and cons of each approach!

matthewbrown77:
Closes #159


package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "dev": "next",
     "build": "next build",
+    "debug": "NODE_OPTIONS='--inspect' next",
```
matthewbrown77, line 8: I added this yarn `debug` script that I thought might be helpful to other backend developers. It runs the app with a remote debugger that you can either attach to from vscode or visit using chrome.

.gitignore
```diff
@@ -18,6 +18,7 @@
 # misc
 .DS_Store
 .idea
+.vscode
```
matthewbrown77, line 21: Added to ignore my debug config for vscode. Also may be helpful for other devs.

pages/events/[id].tsx
```diff
@@ -0,0 +1,60 @@
...
+import urls from "config";
+import { getAllEventIds, getEventById } from "server/actions/events";
+
+const NonprofitEventPage: NextPage<InferGetStaticPropsType<
+  typeof getStaticProps
+>> = props => {
+  const router = useRouter();
+
+  if (router.isFallback) {
+    return <div>Loading...</div>;
```
Comment (PC98) [inline pages/events/[id].tsx:20]: can you either add a `TODO` or return `null` for now? We will probably need a designer approved loading page.

pages/events/[id].tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    return {
+      props: {
+        event
+      },
+      revalidate: 1
+    };
+  } catch (error) {
+    return {
+      props: {
```
Comment (PC98) [inline pages/events/[id].tsx:53]: I am not sure but do you think we should add `revalidate: 1` here as well?

PC98 (approved):
You will have to resolve conflicts before merging :)

pages/events/[id].tsx
```diff
@@ -0,0 +1,60 @@
...
+  NextPage,
+  GetStaticProps,
+  GetStaticPaths,
+  GetStaticPropsContext,
+  InferGetStaticPropsType
+} from "next";
+
+import { useRouter } from "next/router";
+import ErrorPage from "next/error";
+import urls from "config";
```
Comment (PC98) [inline pages/events/[id].tsx:11]: Also this is a nit but we are encouraging everyone to just use `import config from "config"` consistently to avoid confusion about the imported object

pages/events/[id].tsx
```diff
@@ -0,0 +1,60 @@
...
+  return null;
+};
+
+export const getStaticPaths: GetStaticPaths = async () => {
+  const ids = await getAllEventIds();
+
+  return { paths: ids.map(urls.pages.event), fallback: true };
+};
+
+export const getStaticProps: GetStaticProps = async (
```
Comment (PC98) [inline pages/events/[id].tsx:37]: You might also have to drop `GetStaticProps` here, see discussion: https://github.com/GTBitsOfGood/gen-soln/issues/83#issuecomment-697052497

pages/events/[id].tsx
Comment (matthewbrown77) [reply pages/events/[id].tsx:53]: From testing both cases, it looks like the page still attempts to regenerate even if the `revalidate` option is not set in the error case.

pages/events/[id].tsx
Comment (matthewbrown77) [reply pages/events/[id].tsx]: Dropped in 47f7ee8

pages/events/[id].tsx
Comment (matthewbrown77) [reply pages/events/[id].tsx]: Changed in 47f7ee8

pages/events/[id].tsx
Comment (matthewbrown77) [reply pages/events/[id].tsx]: Replaced with null in 47f7ee8

utils/types.ts
```diff
@@ -34,8 +34,9 @@ export interface Donation {
 // Keep in sync with the backend schema
 interface EventBase {
   name: string;
-  startDate: Date;
-  endDate: Date;
+  nonprofitId: string;
```
PC98, line 37: @matthewbrown77 could you remove `nonprofitId` from here? Seems like while resolving conflicts you mistakenly added it and it's failing type check

matthewbrown77: My mistake. Should be fixed now.

PC98: No worries :)

PR #184, opened 2020-09-23
Regenerate donate page if the nonprofit id doesn't exist 

### LINKED ISSUE #83: [DMS] Regenerate donate page if the nonprofit id doesn't exist

ML-Chen:
Fixes #83 
Tested it with making new nonprofits in MongoDB Compass (with a string _id field) and it seems to work properly.

### DIFF
--- a/pages/donate/[id].tsx
+++ b/pages/donate/[id].tsx
@@ -1,25 +1,19 @@
 import React from "react";
 import {
   NextPage,
-  GetStaticProps,
   GetStaticPaths,
   InferGetStaticPropsType,
   GetStaticPropsContext
 } from "next";
-
+import ErrorPage from "next/error";
 import { Dropdown } from "utils/types";
-
 import {
   getNonprofitIds,
   getNonprofitNamesWithIds,
   getNonprofitById
 } from "server/actions/nonprofit";
-
 import DonationPage from "components/donation/DonationPage";
-
 import config from "config";
-import Head from "next/head";
-import ErrorPage from "next/error";
 
 const NonprofitDonationPage: NextPage<InferGetStaticPropsType<
   typeof getStaticProps
@@ -65,7 +59,8 @@ export const getStaticProps = async (context: GetStaticPropsContext) => {
     };
   } catch (err) {
     return {
-      props: {}
+      props: {},
+      revalidate: 10
     };
   }
 };
--- a/pages/donate/[id].tsx
+++ b/pages/donate/[id].tsx
@@ -14,11 +14,15 @@ import {
 } from "server/actions/nonprofit";
 import DonationPage from "components/donation/DonationPage";
 import config from "config";
+import { useRouter } from "next/router";
 
 const NonprofitDonationPage: NextPage<InferGetStaticPropsType<
   typeof getStaticProps
 >> = props => {
-  if (props.nonprofit) {
+  const router = useRouter();
+  if (router.isFallback) {
+    return null;
+  } else if (props.nonprofit) {
     return <DonationPage {...props} />;
   } else {
     return <ErrorPage statusCode={404} />;
@@ -31,6 +35,7 @@ export const getStaticPaths: GetStaticPaths = async () => {
   return { paths: ids.map(config.pages.donate), fallback: true };
 };
 
+// eslint-disable-next-line
 export const getStaticProps = async (context: GetStaticPropsContext) => {
   const id = context.params?.id as string;
 
--- a/pages/donate/[id].tsx
+++ b/pages/donate/[id].tsx
@@ -19,24 +19,15 @@ import DonationPage from "components/donation/DonationPage";
 
 import config from "config";
 import Head from "next/head";
-import DefaultErrorPage from "next/error";
+import ErrorPage from "next/error";
 
 const NonprofitDonationPage: NextPage<InferGetStaticPropsType<
   typeof getStaticProps
 >> = props => {
   if (props.nonprofit) {
-    // @ts-ignore temp
     return <DonationPage {...props} />;
   } else {
-    return (
-      <>
-        <Head>
-          <meta name="robots" content="noindex" />
-          <title>404</title>
-        </Head>
-        <DefaultErrorPage statusCode={404} />
-      </>
-    );
+    return <ErrorPage statusCode={404} />;
   }
 };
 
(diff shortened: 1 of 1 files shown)


pages/donate/[id].tsx
```diff
@@ -1,52 +1,68 @@
...
 
-const NonprofitDonationPage: NextPage<React.ComponentProps<
-  typeof DonationPage
->> = props => <DonationPage {...props} />;
+const NonprofitDonationPage: NextPage<InferGetStaticPropsType<
+  typeof getStaticProps
+>> = props => {
+  if (props.nonprofit) {
+    return <DonationPage {...props} />;
+  } else {
```
Comment (PC98) [inline pages/donate/[id].tsx:23]: nit: don't need the `else {`

pages/donate/[id].tsx
```diff
@@ -1,52 +1,68 @@
...
-
 import config from "config";
 
-const NonprofitDonationPage: NextPage<React.ComponentProps<
-  typeof DonationPage
->> = props => <DonationPage {...props} />;
+const NonprofitDonationPage: NextPage<InferGetStaticPropsType<
+  typeof getStaticProps
+>> = props => {
+  if (props.nonprofit) {
```
Comment (PC98) [inline pages/donate/[id].tsx:21]: Don't forget to check for fallback! See: https://github.com/GTBitsOfGood/gen-soln/blob/develop/pages/events/%5Bid%5D.tsx#L18

pages/donate/[id].tsx
```diff
@@ -1,52 +1,68 @@
...
 export const getStaticPaths: GetStaticPaths = async () => {
   const ids = await getNonprofitIds();
 
-  return { paths: ids.map(config.pages.donate), fallback: false };
+  return { paths: ids.map(config.pages.donate), fallback: true };
 };
 
-export const getStaticProps: GetStaticProps = async ({ params }) => {
-  const id = params?.id as string;
+export const getStaticProps = async (context: GetStaticPropsContext) => {
```
Comment (PC98) [inline pages/donate/[id].tsx:34]: Nit: add an eslint-disable-next-line to skip this warning

pages/donate/[id].tsx
```diff
@@ -1,52 +1,68 @@
...
+      })
+    );
+
+    return {
+      props: {
+        nonprofit,
+        items,
+        selectedValue: id
+      },
+      revalidate: 10
```
Comment (PC98) [inline pages/donate/[id].tsx:58]: Question: do we know how to pick a value for revalidate? Is it just try and see what happens on prod? I think for `events/[id].tsx` we just picked 1 since that's what the example of Next.js' docs uses.

Comment (ML-Chen) [reply pages/donate/[id].tsx:58]: I don't know

pages/donate/[id].tsx
Comment (ML-Chen) [reply pages/donate/[id].tsx:23]: Personally I prefer the explicit `else {` unless one branch of the if-else is much longer.

pages/donate/[id].tsx
Comment (ML-Chen) [reply pages/donate/[id].tsx:21]: Is checking `props.nonprofit` not sufficient?

pages/donate/[id].tsx
Comment (PC98) [reply pages/donate/[id].tsx:58]: ah okay it's fine then

PC98 (approved):

PR #188, opened 2020-09-24
Add UI and functionality to clear filters

### LINKED ISSUE #155: [VMS] Build functionality to clear all selected filters
* As shown in the [Figma designs](https://www.figma.com/file/HMEsyyl6EGJ4hMaGBNtQNe/Donor-and-Volunteer-Experience-Fall20?node-id=0%3A1), whenever a filter is selected, the orange "CLEAR (num)" button appears next to the "Filters" text.
* Implement the UI for this button.
* Add functionality to it; since we are using the URL to store, read and write our filter state, pressing the button should essentially remove all query parameters. Finding the (num) of filters selected is also easy -- you can use the URL query parameters to determine how many filters are being used. Feel free to use `useRouter` directly, I am not sure if you will be able to use the hook created in #181 for this.

migueldlr:
Closes #155 

This change counts the number of filters applied and adds it to the top bar as a button:

https://streamable.com/76x5b6

And it's keyboard-focusable!

https://streamable.com/42f5ze

### DIFF
--- a/components/events/EventsPageLeftRailComponent.tsx
+++ b/components/events/EventsPageLeftRailComponent.tsx
@@ -3,11 +3,11 @@ import { useRouter } from "next/router";
 
 import makeStyles from "@material-ui/core/styles/makeStyles";
 import Typography from "@material-ui/core/Typography";
+import Button from "@material-ui/core/Button";
 import EventsPageLeftRailFilter from "./EventsPageLeftRailFilter";
 import EventsPageTimeFilter from "./EventsPageTimeFilter";
 
 import EventsPageLocationFilterAutocompleteInput from "./EventsPageLocationFilterAutocompleteInput";
-import FocusVisibleOnly from "components/FocusVisibleOnly";
 
 const useStyles = makeStyles({
   root: {
@@ -27,7 +27,8 @@ const useStyles = makeStyles({
     display: "flex",
     flexDirection: "row",
     justifyContent: "space-between",
-    alignItems: "center"
+    alignItems: "center",
+    flexWrap: "wrap"
   },
   clearFilterLabel: {
     // TODO: replace this with a Typography component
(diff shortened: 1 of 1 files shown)


components/events/EventsPageLeftRailComponent.tsx
```diff
@@ -20,14 +22,58 @@ const useStyles = makeStyles({
...
+    });
+  };
+
   return (
     <div className={root}>
-      <Typography className={header}>Filters</Typography>
+      <div className={topBar}>
+        <Typography className={header}>Filters</Typography>
+        {filterCount > 0 && (
+          <FocusVisibleOnly onClick={() => clearFilters()}>
```
PC98, line 70: Why not just `onClick={clearFilters}`?

components/events/EventsPageLeftRailComponent.tsx
```diff
@@ -20,14 +22,58 @@ const useStyles = makeStyles({
...
+            <Typography className={clearFilterLabel}>
+              Clear ({filterCount})
+            </Typography>
+          </FocusVisibleOnly>
```
PC98, line 74: Can we change the underlying component to this: https://material-ui.com/components/buttons/#text-buttons? It seems a bit hacky to use a `Typography` component surrounded by a clickable `div` to achieve this UI. 
I talked to the designers and they are okay with the hover effect that comes with using a Text Button.

components/events/EventsPageLeftRailComponent.tsx
```diff
@@ -20,14 +22,58 @@ const useStyles = makeStyles({
...
   }
 });
 
 const EventsPageLeftRailComponent: React.FC<Record<string, unknown>> = () => {
-  const { root, header } = useStyles();
+  const { root, header, topBar, clearFilterLabel } = useStyles();
+  const router = useRouter();
+
+  // Sum up all the applied filters
+  const filterCount = Object.values(router.query).reduce((acc, x) => {
```
PC98, line 49: nit: maybe we should add a TODO here about only counting query parameters whose names match the filters we support. For instance, if for whatever reason there is a `/?foo=bar&time=TODAY` in our query, `bar` will also get counted in our sum here, even though we have no filter corresponding to `foo`.

PC98 (approved):
🔥

PR #190, opened 2020-09-24
added causes to nonprofit schema

schen2425:
Added causes field to nonprofit schema
NOTE: some of the ENUM  may not be correct

### DIFF
--- a/server/models/nonprofit.js
+++ b/server/models/nonprofit.js
@@ -5,6 +5,18 @@ const Schema = mongoose.Schema;
 const ALPHABET = "0123456789";
 const ID_LENGTH = 8;
 const nanoid = customAlphabet(ALPHABET, ID_LENGTH);
+const CAUSES = [
+  { text: "arts", value: "ARTS" },
+  { text: "Culture and Humanities", value: "CULTURE AND HUMANITIES" },
+  { text: "Education and Research", value: "EDUCATION AND RESEARCH" },
+  { text: "Environment and Animals", value: "ENVIRMONMENT AND ANIMALS" },
+  { text: "Health", value: "HEALTH" },
+  { text: "Human Services", value: "HUMAN SERVICES" },
+  { text: "International", value: "INTERNATIONAL" },
+  { text: "Public, Societal", value: "PUBLIC, SOCIETAL" },
+  { text: "Religion", value: "RELIGION" },
+  { text: "Other", value: "OTHER" }
+];
 
 // Keep in sync with utils/types Nonprofit
 const nonprofitSchema = new Schema({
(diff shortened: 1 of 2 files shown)


server/models/nonprofit.js
```diff
@@ -5,6 +5,18 @@ const Schema = mongoose.Schema;
 const ALPHABET = "0123456789";
 const ID_LENGTH = 8;
 const nanoid = customAlphabet(ALPHABET, ID_LENGTH);
+const CAUSES = [
+  { text: "arts", value: "ARTS" },
```
PC98, line 9: `text: "Arts"`

server/models/nonprofit.js
```diff
@@ -5,6 +5,18 @@ const Schema = mongoose.Schema;
...
+  { text: "Culture and Humanities", value: "CULTURE AND HUMANITIES" },
+  { text: "Education and Research", value: "EDUCATION AND RESEARCH" },
+  { text: "Environment and Animals", value: "ENVIRMONMENT AND ANIMALS" },
+  { text: "Health", value: "HEALTH" },
+  { text: "Human Services", value: "HUMAN SERVICES" },
+  { text: "International", value: "INTERNATIONAL" },
+  { text: "Public, Societal", value: "PUBLIC, SOCIETAL" },
+  { text: "Religion", value: "RELIGION" },
+  { text: "Other", value: "OTHER" }
```
PC98, line 18: Can we use the underscore character to separate the words in "value"? So for instance, `CULTURE AND HUMANITIES` would become `CULTURE_AND_HUMANITIES`. Also as an FYI, you have a typo in "ENVIRMONMENT".

server/models/nonprofit.js
```diff
@@ -62,8 +74,25 @@ const nonprofitSchema = new Schema({
...
+      "ARTS, CULTURE, AND HUMANITIES",
+      "EDUCATION AND RESEARCH",
+      "ENVIRONMENT AND ANIMALS",
+      "HEALTH",
+      "HUMAN SERVICES",
+      "INTERNATIONAL",
+      "PUBLIC, SOCIETAL",
+      "RELIGION",
+      "OTHER"
+    ],
```
PC98, line 90: To avoid having to maintain the same list in two places, you can just do:
```
enum: CAUSES.map(cause => cause.value)
```

server/models/nonprofit.js
```diff
@@ -62,8 +74,25 @@ const nonprofitSchema = new Schema({
...
+    default: "OTHER",
```
PC98, line 91: Hm, I am thinking if there is a better way to specify the default instead of having `"OTHER"` in two places. How about?
```javascript
const DEFAULT_CASE = { text: "Other", value: "OTHER" }
const CAUSES = [
  // stuff
  DEFAULT_CASE
]
// stuff
default: DEFAULT_CASE.value
```
Or more simply:
```javascript
default: CAUSES[CAUSES.length - 1].value
```
Feel free to come up with your own approach!

PC98 (approved):

PR #192, opened 2020-09-25
added router query params state, added the functionality to time filter

### LINKED ISSUE #181: [VMS] Create a hook to read, put and remove values from `router.query`
* All of our filters on the /events page will use the URL parameters to read their current state and update the URL parameters when the user interacts with the filters. To understand what this means, have a look at the Time filter demo: https://streamable.com/8on414 Notice how the URL is updated when a filter selection is made.
* Time filter does this via the code defined in lines 61 to 74: https://github.com/GTBitsOfGood/gen-soln/blob/develop/components/events/EventsPageTimeFilter.tsx#L61
* We will need this functionality for Location and Causes filter as well, so let's create a hook to do this for us. I imagine the hook will look like this, but feel free to make your own changes/improvements!
```javascript
import { useCallback, useMemo } from "react";
import { useRouter } from "next/router";

// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
const useRouterQueryParamsState = (key: string) => {
  const router = useRouter();

  const currentState = // follow what the code in `EventsPageTimeFilter.tsx` does to generate `const times`, except you will use `router.query[key]` instead of `router.query.time`
    
  const routerPushWithUpdatedState = useCallback(
    (updatedState: string[]) => {
      void router.push({
        pathname: router.pathname,
        query: { ...router.query, [key]: updatedState }
      });
    },
    [router, key]
  );

  const put = useCallback(
    (value: string) => {
      const updated = [...currentState, value];

      routerPushWithUpdatedState(updated);
    },
    [routerPushWithUpdatedState, currentState]
  );

  const remove = useCallback(
    (value: string) => {
      const updated = currentState.filter(s => s !== value);

      routerPushWithUpdatedState(updated);
    },
    [routerPushWithUpdatedState, currentState]
  );

  return { currentState, put, remove };
};

export default useRouterQueryParamsState;
```
* Then, refactor the code in `EventsPageTimeFilter.tsx` to use this hook. Particularly think about how this line will change: https://github.com/GTBitsOfGood/gen-soln/blob/develop/components/events/EventsPageTimeFilter.tsx#L99 when we use this hook.

jade-law:
Closes #181 
Added router.query hook to events as useRouterQueryParamsState.tsx. Changed time filter to use this hook.

### DIFF
--- a/components/events/EventsPageTimeFilter.tsx
+++ b/components/events/EventsPageTimeFilter.tsx
@@ -57,7 +57,6 @@ const timeOptions: Dropdown[] = [
 ];
 
 const EventsPageTimeFilter: React.FC = () => {
-  const router = useRouter();
   const { option, optionRoot, optionLabel, optionLabelSelected } = useStyles();
   const { currentState, put, remove } = useRouterQueryParamsState("time");
 
--- a/components/events/EventsPageTimeFilter.tsx
+++ b/components/events/EventsPageTimeFilter.tsx
@@ -61,25 +61,10 @@ const EventsPageTimeFilter: React.FC = () => {
   const { option, optionRoot, optionLabel, optionLabelSelected } = useStyles();
   const { currentState, put, remove } = useRouterQueryParamsState("time");
 
-  const times =
-    router.query.time == null
-      ? []
-      : !Array.isArray(router.query.time)
-      ? [router.query.time]
-      : router.query.time;
-
-  const toggle = async (selected: string) => {
-    await router.push({
-      query: times.includes(selected)
-        ? { time: times.filter(s => s !== selected) }
-        : { time: [...times, selected] }
-    });
-  };
-
   return (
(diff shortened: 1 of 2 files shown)


components/events/EventsPageTimeFilter.tsx
```diff
@@ -57,6 +59,7 @@ const timeOptions: Dropdown[] = [
 const EventsPageTimeFilter: React.FC = () => {
   const router = useRouter();
   const { option, optionRoot, optionLabel, optionLabelSelected } = useStyles();
+  const { currentState, put, remove } = useRouterQueryParamsState("time");
 
   const times =
```
PC98, line 64: Can you delete the `times` array and the `toggle` function? We should use `currentState` instead of `times` and I don't think we need `toggle` anymore

PC98 (approved):

PR #196, opened 2020-09-27
Support pagination in event scripts

Fattimo:
initial commit. Support for pagination added. created method that returns number of events within following week of a date

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -6,7 +6,7 @@ import errors from "utils/errors";
 import {
   Event as EventType,
   EventCardData as EventCardDataType,
-  PaginatedEventCards as PaginatedEventCardsDataType
+  PaginatedEventCards as PaginatedEventCardsInterface
 } from "utils/types";
 
 const cardFields: Record<keyof EventCardDataType, 1> = {
@@ -53,7 +53,7 @@ export async function getUpcomingEventsCardData({
     totCount: totCount,
     date: date,
     isLastPage: totCount - page * cardsPerPage < 0
-  } as PaginatedEventCardsDataType;
+  } as PaginatedEventCardsInterface;
 }
 
 export async function getUpcomingEventsCardDataCount(date: Date) {
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -8,7 +8,7 @@ import {
   PaginatedEventCards as PaginatedEventCardsInterface
 } from "utils/types";
 
-const cardFields: Record<keyof EventCardDataType, 1> = {
+const CARD_FIELDS: Record<keyof EventCardDataType, 1> = {
   name: 1,
   startDate: 1,
   endDate: 1,
@@ -17,54 +17,54 @@ const cardFields: Record<keyof EventCardDataType, 1> = {
   nonprofitID: 1,
   duration: 1
 };
-const cardsPerPage = 4;
+const CARDS_PER_PAGE = 4;
+const MILLISECONDS_IN_WEEK = 7 * 24 * 60 * 60 * 1000;
 
-interface pageInformation {
+interface PageInformation {
   date: string;
   page: number;
-  totCount: number;
+  totalCount: number;
 }
 
 export async function getUpcomingEventsCardData({
-  date = new Date().toJSON(),
-  page = 0,
-  totCount
-}: pageInformation) {
+  date,
+  page,
+  totalCount
+}: PageInformation): Promise<PaginatedEventCardsInterface> {
   await Mongo();
 
   const result = await Event.find(
     {
       startDate: {
-        $gte: new Date()
+        $gte: new Date(date),
+        $leq: new Date(new Date(date).getTime() + MILLISECONDS_IN_WEEK)
       }
     },
-    cardFields
+    CARD_FIELDS
   )
     .populate("nonprofitId", "name", Nonprofit)
     .sort({ startDate: 1 })
-    .limit(totCount)
-    .skip(page > 0 ? (page - 1) * cardsPerPage : 0)
-    .limit(cardsPerPage);
+    .skip(page > 0 ? (page - 1) * CARDS_PER_PAGE : 0)
+    .limit(CARDS_PER_PAGE);
 
   return {
     eventCards: result.map(r => r.toJSON()) as EventCardDataType[],
-    page: page,
-    totCount: totCount,
-    date: date,
-    isLastPage: totCount - page * cardsPerPage < 0
-  } as PaginatedEventCardsInterface;
+    page,
+    totalCount,
+    date,
+    isLastPage: totalCount - (page + 1) * CARDS_PER_PAGE <= 0
+  };
 }
 
 export async function getUpcomingEventsCardDataCount(date: Date) {
   await Mongo();
 
-  const result = await Event.countDocuments({
+  return Event.countDocuments({
     startDate: {
       $gte: date,
-      $lte: new Date(date.getMilliseconds() + 6.048e8)
+      $lte: new Date(date.getTime() + MILLISECONDS_IN_WEEK)
     }
   });
-  return result;
 }
 
 interface Coordinates {
@@ -75,7 +75,7 @@ interface Coordinates {
 export async function getNearestEventsCardData({ lat, long }: Coordinates) {
   await Mongo();
 
-  const result = await Event.find({}, cardFields)
+  const result = await Event.find({}, CARD_FIELDS)
     .populate("nonprofitId", "name", Nonprofit)
     .where("address.location")
     .near({
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -1,4 +1,3 @@
-import { DatePickerProps } from "material-ui";
 import Mongo from "server/index";
 import Event from "server/models/event";
 import Nonprofit from "server/models/nonprofit";
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -1,10 +1,12 @@
+import { DatePickerProps } from "material-ui";
 import Mongo from "server/index";
 import Event from "server/models/event";
 import Nonprofit from "server/models/nonprofit";
 import errors from "utils/errors";
 import {
   Event as EventType,
-  EventCardData as EventCardDataType
+  EventCardData as EventCardDataType,
+  PaginatedEventCards as PaginatedEventCardsDataType
 } from "utils/types";
 
 const cardFields: Record<keyof EventCardDataType, 1> = {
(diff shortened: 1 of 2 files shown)


utils/types.ts
```diff
@@ -51,6 +51,14 @@ export type EventCardData = EventBase & {
   nonprofitID: Pick<Nonprofit, "_id" | "name">;
 };
 
+export type PaginatedEventCards = {
+  eventCards: [EventCardData];
+  page: number;
+  totCount: number;
+  date: string;
+  isLastPage: boolean;
+};
```
Fattimo: > Now we should modify what getUpcomingEventsCardData returns. It should not only return an array of Event documents but also the passed-in date, passed-in page and a boolean value: isLastPage which will either be true or false based on the passed in totalCount, the current page number and our nPerPage=4

`nPerPage` is referenced here, but I did not specify it as a field in the interface* (will change to interface in next commit), as it was not specified as explicitly being needed in the original method header. I do think `nPerPage` should be in both this interface as well as in the method header, so as to support dynamic UI updates from the client window resizing

server/actions/events.ts
```diff
@@ -30,9 +43,29 @@ export async function getUpcomingEventsCardData() {
...
+  return {
+    eventCards: result.map(r => r.toJSON()) as EventCardDataType[],
+    page: page,
+    totCount: totCount,
+    date: date,
+    isLastPage: totCount - page * cardsPerPage < 0
+  } as PaginatedEventCardsDataType;
+}
+
+export async function getUpcomingEventsCardDataCount(date: Date) {
```
Fattimo, line 58: Should this take type `Date` or should we use a `string`, as in `getUpcomingEventCardsData()`?

server/actions/events.ts
```diff
@@ -30,9 +43,29 @@ export async function getUpcomingEventsCardData() {
...
+  await Mongo();
+
+  const result = await Event.countDocuments({
+    startDate: {
+      $gte: date,
+      $lte: new Date(date.getMilliseconds() + 6.048e8)
+    }
+  });
+  return result;
```
Fattimo, line 67: return await

server/actions/events.ts
```diff
@@ -16,8 +18,19 @@ const cardFields: Record<keyof EventCardDataType, 1> = {
...
+interface pageInformation {
+  date: string;
+  page: number;
+  totCount: number;
+}
+
+export async function getUpcomingEventsCardData({
+  date = new Date().toJSON(),
+  page = 0,
+  totCount
```
Fattimo, line 31: See bottom comment on `nPerPage`, but I think that it would be good to have `nPerPage` as a parameter, specifically in the case that the screen will dynamically update the amount of cards in a slider deck based on window sizing.

Additionally, my main concern was on the usage of `totCount`. It is referenced that 

> totalCount (a number, this will be the result of calling getUpcomingEventsCardDataCount)

The default value of this variable can be set to the result of calling this method, which I think would be the best option, but the current issue is that the method returns a `Promise`, which can't be used in a method argument.

The alternative, which this method currently supports and is less computationally expensive is to call `getUpcomingEventsCardDataCount()` from the client-side (which stores this number on the client-side), and this number can be piped into these server actions

server/actions/events.ts
```diff
@@ -30,9 +42,29 @@ export async function getUpcomingEventsCardData() {
...
+  } as PaginatedEventCardsInterface;
+}
+
+export async function getUpcomingEventsCardDataCount(date: Date) {
+  await Mongo();
+
+  const result = await Event.countDocuments({
+    startDate: {
+      $gte: date,
+      $lte: new Date(date.getMilliseconds() + 6.048e8)
```
PC98, line 64: I think you meant `date.getTime()`, which will give you milliseconds since the UNIX epoch. You can reach the difference between getTime and getMilliseconds here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

Also, can you declare a constant `const MILLISECONDS_IN_A_WEEK = 7 * 24 * 60 * 60 *  1000` so that it's more clear how you go that `6.048e8`

utils/types.ts
```diff
@@ -51,6 +51,14 @@ export type EventCardData = EventBase & {
   nonprofitID: Pick<Nonprofit, "_id" | "name">;
 };
 
+export interface PaginatedEventCards {
+  eventCards: [EventCardData];
```
PC98, line 55: Not sure if `[EventCardData]` is correct; use `EventCardData[]` or `Array< EventCardData>`

server/actions/events.ts
```diff
@@ -16,8 +17,19 @@ const cardFields: Record<keyof EventCardDataType, 1> = {
   nonprofitID: 1,
   duration: 1
 };
+const cardsPerPage = 4;
```
PC98, line 20: nit: uppercase constants, `CARDS_PER_PAGE` (even `cardFields` you can make uppercase).

server/actions/events.ts
```diff
@@ -16,8 +17,19 @@ const cardFields: Record<keyof EventCardDataType, 1> = {
...
-export async function getUpcomingEventsCardData() {
+interface pageInformation {
+  date: string;
+  page: number;
+  totCount: number;
+}
+
+export async function getUpcomingEventsCardData({
+  date = new Date().toJSON(),
+  page = 0,
```
PC98, line 30: there is no point of a default value I think since in the above interface `date`, and `page` are required fields

server/actions/events.ts
PC98: Yep it makes sense to pass this information to the client so that it doesn't have to be re-computed again.

>  but the current issue is that the method returns a Promise, which can't be used in a method argument.

As far this is concerned, this is how I imagine the first call to `getUpcomingEventsCardData` to work:
```javascript
// pages/events/[id].tsx
export const getServerSideProps: GetServerSideProps = async (context) => {
  // stuff to check if context.query is empty
  const date = new Date();
  const upcomingEventsTotalCount = await getUpcomingEventsCardDataCount(date);
  const upcomingEventsFirstPageData = await getUpcomingEventsCardData({ date, page: 0, totCount: upcomingEventsTotalCount })
  // return stuff
};
```

server/actions/events.ts
```diff
@@ -16,8 +17,19 @@ const cardFields: Record<keyof EventCardDataType, 1> = {
   nonprofitID: 1,
   duration: 1
 };
+const cardsPerPage = 4;
 
-export async function getUpcomingEventsCardData() {
+interface pageInformation {
```
PC98, line 22: Nit: begin with a capital letter, `PageInformation`.

server/actions/events.ts
```diff
@@ -16,8 +17,19 @@ const cardFields: Record<keyof EventCardDataType, 1> = {
...
+  date: string;
+  page: number;
+  totCount: number;
```
PC98, line 25: nit: `totalCount`

server/actions/events.ts
```diff
@@ -30,9 +42,29 @@ export async function getUpcomingEventsCardData() {
   )
     .populate("nonprofitId", "name", Nonprofit)
     .sort({ startDate: 1 })
-    .limit(5);
+    .limit(totCount)
```
PC98, line 45: I think you should instead modify your query -- add the `$lte: new Date(date.getMilliseconds() + 6.048e8)` here as well, instead of having a `.limit(totCount)`.

Also don't forget to use `$gte: new Date(date)` here! That's the entire point of passing a `date` to this function; otherwise `$gte: new Date()` will change our filter each time this function is called.

server/actions/events.ts
```diff
@@ -30,9 +42,29 @@ export async function getUpcomingEventsCardData() {
...
+    .skip(page > 0 ? (page - 1) * cardsPerPage : 0)
+    .limit(cardsPerPage);
 
-  return result.map(r => r.toJSON()) as EventCardDataType[];
+  return {
+    eventCards: result.map(r => r.toJSON()) as EventCardDataType[],
+    page: page,
+    totCount: totCount,
+    date: date,
```
PC98, line 53: one really nice thing about JS syntax:
```
eventCards: // whatever
page,
totCount,
date,
isLastPage: // whatever

File: server/actions/events.ts [hunk 13]
```diff
@@ -30,9 +42,29 @@ export async function getUpcomingEventsCardData() {
...
+    isLastPage: totCount - page * cardsPerPage < 0
+  } as PaginatedEventCardsInterface;
```
PC98, line 55: Instead of `as PaginatedEventCardsInterface` here, specify the return type as `Promise<PaginatedEventCardsInterface>` in your function definition.

In general, we should avoid using `as`. We are using it for `EventCardDataType` because our Mongoose models aren't typed with TypeScript.

server/actions/events.ts
```diff
@@ -30,9 +42,29 @@ export async function getUpcomingEventsCardData() {
...
+    .skip(page > 0 ? (page - 1) * cardsPerPage : 0)
+    .limit(cardsPerPage);
 
-  return result.map(r => r.toJSON()) as EventCardDataType[];
+  return {
+    eventCards: result.map(r => r.toJSON()) as EventCardDataType[],
+    page: page,
+    totCount: totCount,
+    date: date,
+    isLastPage: totCount - page * cardsPerPage < 0
```
PC98, line 54: `totCount - (page + 1) * cardsPerPage < 0` perhaps?

For instance, if `totCount` is  7 and `cardsPerPage` is 4, we expect `page=0` to contain the first 4 elements and `page=1` to contain the remaining 3. `page=1` would be the last page here.

server/actions/events.ts
PC98: either way is fine 👍

server/actions/events.ts
PC98: For this:

> I think that it would be good to have nPerPage as a parameter, specifically in the case that the screen will dynamically update the amount of cards in a slider deck based on window sizing.

I agree but lets things simple for now and use a fixed value in this PR, we will have to talk to the front-end devs about how the UI is being implemented and then proceed with any future changes.

server/actions/events.ts
Fattimo: Yep, TS infers Promise<number>

server/actions/events.ts
Fattimo: My original implementation had this, but I figured that if the client side is already calculating the number of documents within a week (https://github.com/GTBitsOfGood/gen-soln/pull/196#discussion_r495622321), then there is no need to recalculate the total number of documents again -- we just need to pull the first `totalCount` documents we find from `date`

PC98 (approved):

PR #201, opened 2020-09-29
Jade/96 location filters

### LINKED ISSUE #96: [VMS] Populate the location filter with results from autocomplete text component and modify page URL
* After #94 and #95 done, it's time to emulate the behavior that one can observe here: https://careers.google.com/jobs/results
* Go to the above page, use the "Locations" filter to search for one or more cities. Notice how the selected cities appear as [`Chip`s](https://material-ui.com/components/chips/#chip) above the location input, and the URL updates to include multiple `location` query parameters.
* That's what you will do for this task on our app! You are not going to write any filtering code right now, but just implement the UI of having `Chip`s appear **below** the auto-complete location input (implemented in #95) that is part of the filtering container (done in #94). The `Chip`s should have a **button to delete/remove them.** Helpful demo: https://material-ui.com/components/autocomplete/#multiple-values
* Be sure to use the hook created in #181 for reading/writing state from URL query parameters! For that hook, you can pass the `key` value as `"location"`.

jade-law:
Closes #96 
Added chips for results of location filter. Linked to the hook created in #181 .

### DIFF
--- a/components/events/EventsPageLeftRailComponent.tsx
+++ b/components/events/EventsPageLeftRailComponent.tsx
@@ -6,10 +6,7 @@ import Typography from "@material-ui/core/Typography";
 import Button from "@material-ui/core/Button";
 import EventsPageLeftRailFilter from "./EventsPageLeftRailFilter";
 import EventsPageTimeFilter from "./EventsPageTimeFilter";
-import EventsPageLocationFilterAutocompleteInput from "./EventsPageLocationFilterAutocompleteInput";
 import EventPageLocationFilter from "./EventPageLocationFilter";
-
-import useRouterQueryParamsState from "./useRouterQueryParamsState";
 
 const useStyles = makeStyles({
   root: {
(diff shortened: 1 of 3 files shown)


components/events/EventsPageLeftRailComponent.tsx
```diff
@@ -76,7 +86,15 @@ const EventsPageLeftRailComponent: React.FC<Record<string, unknown>> = () => {
...
-        content={<EventsPageLocationFilterAutocompleteInput />}
+        content={
+          <EventsPageLocationFilterAutocompleteInput
+            addLocationChip={addLocationChip}
+          />
+        }
+      />
+      <EventPageLocationFilter
+        content={currentState}
+        deleteLocation={removeLocationChip}
```
PC98, line 97: Insted of all this, can we just have:
```javascript
      <EventsPageLeftRailFilter
        header="Location"
        content={
          <EventPageLocationFilter />
        }
      />
```

And then `EventPageLocationFilter` can take care of rendering `EventsPageLocationFilterAutocompleteInput` and using the hook `useRouterQueryParamsState("location");`

PC98:
@jade-law I think you misread my comment 😅
I was asking you to use `useRouterQueryParamsState("location")` in `EventPageLocationFilter` but it seems you added it to both `EventPageLocationFilter` and `EventsPageLocationFilterAutocompleteInput`. I will just push a commit to fix this -- your first version was pretty close to what we wanted.

PC98 (approved):

PR #202, opened 2020-09-29
Passed list of causes to EventsPageLeftRailComponent

schen2425:
Passed list of causes to EventsPageLeftRailComponent


server/actions/nonprofit.ts
```diff
@@ -106,3 +106,13 @@ export async function linkStripeAccount(
...
 }
+
+export async function getCauses(): Promise<Dropdown[]> {
+  const Causes = CAUSES.map(
+    ({ text, value }): Dropdown => ({
+      value: value,
+      text: text
+    })
+  );
+  return Causes;
```
PC98, line 117: 1. I think you can remove the "async"
2. Can you just do:
```
export function getCauses(): Dropdown[] {
  return CAUSES 
}
```
If TS complains:
```
export function getCauses() {
  return CAUSES as  Dropdown[]
}
```

components/events/EventsPageLeftRailComponent.tsx
```diff
@@ -42,7 +44,14 @@ const useStyles = makeStyles({
...
 
-const EventsPageLeftRailComponent: React.FC<Record<string, unknown>> = () => {
+interface Props {
+  causesFilterOptions: Dropdown[];
+}
+
+const EventsPageLeftRailComponent: React.FC<Record<
+  string,
+  unknown
+>> = props => {
```
PC98, line 54: should use `React.FC<Props>` here, as you did for `EventsPage`

PC98 (approved):

PC98:
Ah I meant the entire thing should just be `React.FC<Props>`:
```
const EventsPageLeftRailComponent: React.FC<Props>
```
but I will fix this real quick so that our CI check can pass, everything else looks good!

PR #203, opened 2020-09-29
Add pagination to event location queries

### LINKED ISSUE #193: [VMS] Add pagination support to `getNearestEventsCardData()`
* Follow the instructions in #186, the changes for this task will be nearly identical.
* One new product requirement is that `getNearestEventsCardData()` should only return events within 20 miles of the passed in `{ lat, long }` coordinates. Check if the `near` query can be modified to take into account some sort of radius field.

matthewbrown77:
Closes #193

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -24,7 +24,7 @@ const CARD_FIELDS: Record<keyof EventCardDataType, 1> = {
 };
 const CARDS_PER_PAGE = 4;
 const MILLISECONDS_IN_WEEK = 7 * 24 * 60 * 60 * 1000;
-const RADIUS_OF_EARTH = 32186.9;
+const NEAREST_EVENTS_RADIUS = 20 / 3959; // radius for nearest events in radians (20 miles / earth's radius)
 
 export async function getUpcomingEventsCardData({
   date,
@@ -78,7 +78,7 @@ export async function getNearestEventsCardData({
     {
       "address.location": {
         $geoWithin: {
-          $centerSphere: [[location.long, location.lat], 20 / RADIUS_OF_EARTH]
+          $centerSphere: [[location.long, location.lat], NEAREST_EVENTS_RADIUS]
         }
       }
     },
@@ -105,7 +105,9 @@ export async function getNearestEventsCardDataCount({
 
   return Event.countDocuments({
     "address.location": {
-      $geoWithin: { $centerSphere: [[long, lat], 20 / RADIUS_OF_EARTH] }
+      $geoWithin: {
+        $centerSphere: [[long, lat], NEAREST_EVENTS_RADIUS]
+      }
     }
   });
 }
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -2,10 +2,15 @@ import Mongo from "server/index";
 import Event from "server/models/event";
 import Nonprofit from "server/models/nonprofit";
 import errors from "utils/errors";
+
 import {
+  Coordinates,
   Event as EventType,
   EventCardData as EventCardDataType,
-  PaginatedEventCards as PaginatedEventCardsInterface
+  DatePaginatedEventCards,
+  DatePageInformation,
+  LocationPaginatedEventCards,
+  LocationPageInformation
 } from "utils/types";
 
 const CARD_FIELDS: Record<keyof EventCardDataType, 1> = {
@@ -19,18 +24,13 @@ const CARD_FIELDS: Record<keyof EventCardDataType, 1> = {
 };
 const CARDS_PER_PAGE = 4;
 const MILLISECONDS_IN_WEEK = 7 * 24 * 60 * 60 * 1000;
-
-interface PageInformation {
-  date: string;
-  page: number;
-  totalCount: number;
-}
+const RADIUS_OF_EARTH = 32186.9;
 
 export async function getUpcomingEventsCardData({
   date,
   page,
   totalCount
-}: PageInformation): Promise<PaginatedEventCardsInterface> {
+}: DatePageInformation): Promise<DatePaginatedEventCards> {
   await Mongo();
 
   const result = await Event.find(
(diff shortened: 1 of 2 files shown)


utils/types.ts
```diff
@@ -51,14 +51,39 @@ export type EventCardData = EventBase & {
...
+
+interface PaginatedEventCards {
   eventCards: EventCardData[];
   page: number;
   totalCount: number;
-  date: string;
   isLastPage: boolean;
 }
 
+export interface DatePaginatedEventCards extends PaginatedEventCards {
```
matthewbrown77, line 66: To improve reuse for the pagination typings, I had each new interface extend a base interface so the only differing fields will be `location` or `date`. I considered using optional fields, but that seemed to defeat the point of type checking since it could be undefined/

utils/types.ts
```diff
@@ -51,14 +51,39 @@ export type EventCardData = EventBase & {
   nonprofitID: Pick<Nonprofit, "_id" | "name">;
 };
 
-export interface PaginatedEventCards {
+export interface Coordinates {
+  lat: number;
+  long: number;
```
matthewbrown77, line 56: One thing we should look into for next spring is that I think the coordinates in each event should match the query style in mongo. Mongo uses` long, lat`, whereas we are currently using `lat, long`. It makes looking at location data very confusing and inconsistent and I think it has an impact on the query as well. Running the new query for data returned 0 results until I reversed the location coordinates in the db.
event_data

server/actions/events.ts
```diff
@@ -67,24 +67,47 @@ export async function getUpcomingEventsCardDataCount(date: Date) {
...
+  page,
+  totalCount
+}: LocationPageInformation): Promise<LocationPaginatedEventCards> {
+  await Mongo();
+
+  const result = await Event.find(
+    {
+      "address.location": {
+        $geoWithin: {
+          $centerSphere: [[location.long, location.lat], 20 / RADIUS_OF_EARTH]
```
matthewbrown77, line 81: The $centerSphere operator takes a pair of coordinates and a radius in radians. To achieve the desired query, we divide the desired radius in miles by the approximate radius of the earth. See this [mongo reference page](https://docs.mongodb.com/manual/reference/operator/query/centerSphere/#op._S_centerSphere) for more details.

utils/types.ts
PC98: > whereas we are currently using `lat, long`.

Perhaps I am misunderstanding, but I don't think we fixed our order anywhere correct? `interface Coordinates` is like an object type, so the order of the fields is irrelevant. The order does however matter when we are using the coordinates:
```javascript
$geoWithin: { $centerSphere: [[long, lat], 20 / RADIUS_OF_EARTH] }
```
And I think here the order is correct since in our DB we store info as GeoJSON in which longitude comes first.

utils/types.ts
```diff
@@ -51,14 +51,39 @@ export type EventCardData = EventBase & {
...
+  location: Coordinates;
+}
+
+interface PageInformation {
+  page: number;
+  totalCount: number;
+}
+
+export interface DatePageInformation extends PageInformation {
+  date: Date;
```
PC98, line 80: make this type string, we treat `date` as type string so that `getUpcomingEventsCardData` can serialize and pass it to the front-end.

server/actions/events.ts
```diff
@@ -67,24 +67,47 @@ export async function getUpcomingEventsCardDataCount(date: Date) {
...
-    .near({
-      center: [long, lat],
-      spherical: true
-    })
-    .limit(5);
-
-  return result.map(r => r.toJSON()) as EventCardDataType[];
+  return Event.countDocuments({
+    "address.location": {
+      $geoWithin: { $centerSphere: [[long, lat], 20 / RADIUS_OF_EARTH] }
```
PC98, line 108: could you check the math on "20 / RADIUS_OF_EARTH"? This answer https://stackoverflow.com/a/25736795/2212655 seems to suggest using  `some_miles / 3959` since `3959` is Earth's radius in miles. Also, instead of hardcoding the "20", we should either declare a constant for it (as you have for `RADIUS_OF_EARTH`; name it something appropriate like `NEAREST_EVENTS_DISTANCE_THRESHOLD` and add a comment that this is miles) or you can directly compute the `20 / 3959` and store it in an appropriately named variable.

utils/types.ts
```diff
@@ -51,14 +51,39 @@ export type EventCardData = EventBase & {
...
+interface PaginatedEventCards {
   eventCards: EventCardData[];
   page: number;
   totalCount: number;
-  date: string;
   isLastPage: boolean;
 }
 
+export interface DatePaginatedEventCards extends PaginatedEventCards {
+  date: Date;
```
PC98, line 67: Sam as below comment, change this to string

utils/types.ts
```diff
@@ -51,14 +51,39 @@ export type EventCardData = EventBase & {
...
+  totalCount: number;
+}
+
+export interface DatePageInformation extends PageInformation {
+  date: Date;
+}
+
+export interface LocationPageInformation extends PageInformation {
+  location: Coordinates;
+}
```
PC98, line 85: Can we reduce some duplication? Perhaps:
```javascript
interface PageInformation {
  page: number;
  totalCount: number;
  isLastPage: boolean; // seems more idiomatic to have `isLastPage` on this type
}

type PaginatedEventCards = PageInformation & {
  eventCards: EventCardData[];
}

interface PaginateWithLocation {
  location: Coordinates
}

export type LocationPaginatedEventCards = PaginatedEventCards & PaginateWithLocation;

export type LocationPageInformation = PageInformation & PaginateWithLocation;
```
And similarly for date. Feel free to think of alternative ways to reduce the duplication!

utils/types.ts
PC98: Yep, I like this!

server/actions/events.ts
matthewbrown77: You are right, `3959` is the desired value. That was screwing with the results that made me think there was an issue with the order of `lat, long`. The query now returns the correct values.

utils/types.ts
matthewbrown77: I did not know you could do this with types and interfaces. Very cool!

PC98 (approved):
💯

PR #205, opened 2020-09-29
Refactored autocomplete input into shared component

KoolCards:
- Only makes a small update with a new props locationType that is used to specify whether cities or addresses are being searched for.


components/LocationAutocompleteInput.tsx
```diff
@@ -49,6 +53,7 @@ const EventsPageLocationFilterAutocompleteInput: React.FC<Props> = props => {
   const loaded = useRef(false);
 
   if (typeof window !== "undefined" && !loaded.current) {
+    console.log(config.googleMapsKey);
```
PC98, line 56: Please remove this 😅

PC98 (approved):

PR #207, opened 2020-09-29
Create cause filter for events page

### LINKED ISSUE #154: [VMS] Build UI for the Causes filter
* Create a component corresponding to the Causes filter, shown below the Time filter in the [Figma designs](https://www.figma.com/file/HMEsyyl6EGJ4hMaGBNtQNe/Donor-and-Volunteer-Experience-Fall20?node-id=0%3A1).
* See how the Time filter has been implemented: `EventsPageTimeFilter.tsx` and how to use: `EventsPageLeftRailFilter.tsx`: https://github.com/GTBitsOfGood/gen-soln/blob/develop/components/events/EventsPageLeftRailComponent.tsx#L35 for insipration. This will help you implement functionality to minimize the Causes filter.
* Just like the Time filter, we will want to update the URL based on the filter options selected by the user. Task #181 will create a hook that you can use for this purpose.
* **Bonus points if you can figure out a way to share UI code between `EventsPageTimeFilter` and your component.**

matthewbrown77:
Closes #154

### DIFF
--- a/components/events/EventsPageDropdownFilter.tsx
+++ b/components/events/EventsPageDropdownFilter.tsx
@@ -1,3 +1,4 @@
+import React from "react";
 import clsx from "clsx";
 import { Dropdown } from "utils/types";
 
--- a/components/events/EventsPageDropdownFilter.tsx
+++ b/components/events/EventsPageDropdownFilter.tsx
@@ -1,7 +1,6 @@
-import React from "react";
 import clsx from "clsx";
-
 import { Dropdown } from "utils/types";
+
 import {
   Checkbox,
   FormControlLabel,
@@ -12,7 +11,6 @@ import {
 
 import UncheckedIcon from "@horizon/icons/UncheckedIcon";
 import CheckedIcon from "@horizon/icons/CheckedIcon";
-
 import useRouterQueryParamsState from "./useRouterQueryParamsState";
 
 const useStyles = makeStyles({
@@ -39,6 +37,10 @@ const useStyles = makeStyles({
     fontFamily: "Open Sans, sans-serif",
     fontSize: 14,
     lineHeight: "130%",
+    maxWidth: "143px",
+    whiteSpace: "nowrap",
+    overflow: "hidden",
+    textOverflow: "ellipsis",
     color: "#999999"
   },
   optionLabelSelected: {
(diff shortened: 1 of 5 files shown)


matthewbrown77:
This PR refactors the initial `EventsPageTimeFilter` into a more re-usable `EventsPageDropdownFilter`. This filter takes 2 props: the `filter` and `filterOptions`. This way, we can reuse the same code for both filters and pass the filter and options and add additional filters if necessary.

pages/events/index.tsx
```diff
@@ -1,20 +1,38 @@
...
+      timeFilterOptions={props.timeFilterOptions}
+      causesFilterOptions={props.causesFilterOptions}
+    />
+  );
 };
 
 // eslint-disable-next-line @typescript-eslint/require-await
 export const getServerSideProps = async () => {
   // Don't remove the async otherwise InferGetStaticPropsType won't work as expected
+  const timeOptions: Dropdown[] = [
```
matthewbrown77, line 22: We can move `timeOptions` to this method and pass it down to the filters the same way as `causeFilterOptions`.

matthewbrown77:
cause filter

components/events/EventsPageLayout.tsx
```diff
@@ -17,7 +17,7 @@ const useStyles = makeStyles(() =>
       }
     },
     content: {
-      flex: 5,
+      flex: 7,
```
matthewbrown77, line 20: Decided to make the `LeftRailComponent` a little smaller to better match the figma.

pages/events/index.tsx
PC98: We can, but I don't think we need to. Only the front-end uses these values so there is no point sending this from the server on each request. For the causes filters, since our back-end schema had to use the values, we are passing them using `getServerSideProps()` instead of maintaining a copy of the options on both the front-end and the back-end. Though if we can figure out a way to do this without `getServerSideProps()` for causes as well that would be great (I wonder if putting the causes options somewhere in `utils/` and importing them wherever required will suffice 🤔).

components/events/EventsPageDropdownFilter.tsx
```diff
@@ -1,7 +1,6 @@
-import React from "react";
```
PC98, line 1: nit: let's put this back to be consistent

PC98 (approved):

pages/events/index.tsx
matthewbrown77: It's worth having a discussion about it. It could make sense to have a single file in utils contain our dropdown options (but something that could be a task in the next sprint).

components/events/EventsPageDropdownFilter.tsx
matthewbrown77: Put back in the latest commit.

components/events/EventsPageLayout.tsx
migueldlr: I think we should make the left rail a fixed size, actually. On screens from 1024px to 1440px wide, the left rail ends up like 130px wide or so and event cards need ~200px of space (https://www.figma.com/file/HMEsyyl6EGJ4hMaGBNtQNe/Donor-and-Volunteer-Experience-%2F-Fall20?node-id=201%3A1685)

matthewbrown77: I noticed that when messing around with the page on the vercel deployment. I was going to bring it up as a task for next sprint to get the layout of the page ironed out so we can focus on the remaining content.

PR #208, opened 2020-09-29
Create event list

### LINKED ISSUE #156: [VMS] Component for a horizontal list that supports pagination and navigation buttons
* We have to build the list component in the Figma docs that is used for "Upcoming Volunteer Events" and "Volunteer Events Near You".
* This component should be generic enough to support both the above use cases.
* We want to support pagination -- when a user clicks the next navigation button, a function should be called that fetches the items that need to be displayed. While the data is being fetched, we can perhaps show 4 empty cards with glimmers.
* **If you are doing this task, let's schedule a time to talk since I am not entirely sure how we should implement this.**
* Potentially useful links:
    * https://material-ui.com/components/transitions/#slide
    * https://material-ui.com/components/buttons/#icon-buttons
    * https://material-ui.com/components/skeleton/#skeleton
    * https://material-ui.com/components/pagination/#pagination
    * https://material-ui.com/components/tabs/
    * https://github.com/Learus/react-material-ui-carousel
    * https://tj.ie/scrollable-container-controls-with-react/

migueldlr:
Closes #156

`EventsList.tsx` shows a responsive number of event cards (max 4), supporting pagination but not animations.

https://streamable.com/w8sqqy
https://streamable.com/x9rlsc

### Display/Navigation

There are three key pieces of state at play:
- `events`, the array of event information
- `rowSize`, the number of events per row
- `first`, the index of the first element displayed

We want to display elements at indices `[first ... first + rowSize]`. If we still need elements, then we make a request to the server for more items and pad the row with glimmers.  If we don't have any more elements to display (based on `maxElem`), then we don't render any glimmers.

Pressing the previous button moves `first` to `first - rowSize` (or 0 if it becomes negative). Pressing the next button moves `first` to `first + rowSize`.

### Pagination

When we need to show more elements (`first + rowSize > maxElem`), we fetch as many elements as needed to fill the row using `getEvents()`. `getEvents()` is called in an async IIFE. After it is called, we update `events` with the data and set `maxElem` if the server tells us no more elements are available. We also use `fetchingRef` to make sure only one call to the server happens at a time.

### Responsivity

On resize (and on component mount), we grab the size of the container element and do math to figure out how many cards we can fit in given space. We set `rowSize` accordingly (this may or may not trigger a fetch). The resize callback is wrapped in a debounce to make sure we don't update state too much.

Because of the total delay between resize and `rowSize` being updated (debounce + state update), we set the container to wrap and hide the y-overflow so the component retains the expected height at all times.

### Future Work

- Slide animations (React Transition Group will probably be the way to go)
- Pass `getEvents` as a prop so we can reuse the component and not use mock data

### Other Changes
- Add `ChevronLeftIcon` and `ChevronRightIcon` from Horizon's icon set
- Add `EventCardGlimmerLarge` by copying `EventCardLarge` and replacing content with Material UI's `Skeleton` component
- Use `EventList` in a container `div` in `EventsPage`

### DIFF
--- a/components/core/icons/ChevronLeftIcon.tsx
+++ b/components/core/icons/ChevronLeftIcon.tsx
@@ -1,0 +1,26 @@
+import { SvgIcon } from "@material-ui/core";
+import React from "react";
+
+interface Props {
+  color: string;
+}
+
+const ChevronLeftIcon: React.FC<Props> = ({ color }: Props) => {
+  return (
+    <SvgIcon
+      htmlColor={color}
+      width="17"
+      height="17"
+      viewBox="0 0 17 17"
+      fill="none"
+    >
+      <path
+        fillRule="evenodd"
+        clipRule="evenodd"
+        d="M10.0384 13.2803C9.74549 13.5732 9.27061 13.5732 8.97772 13.2803L4.72772 9.03033C4.43483 8.73744 4.43483 8.26256 4.72772 7.96967L8.97772 3.71967C9.27061 3.42678 9.74549 3.42678 10.0384 3.71967C10.3313 4.01256 10.3313 4.48744 10.0384 4.78033L6.31871 8.5L10.0384 12.2197C10.3313 12.5126 10.3313 12.9874 10.0384 13.2803Z"
+      />
+    </SvgIcon>
+  );
+};
+
+export default ChevronLeftIcon;
--- a/components/core/icons/ChevronLeftIcon.tsx
+++ b/components/core/icons/ChevronLeftIcon.tsx
@@ -5,15 +5,9 @@ interface Props {
   color: string;
 }
 
-const ChevronLeftIcon: React.FC<Props> = ({ color }: Props) => {
+const ChevronLeftIcon: React.FC<Props> = ({ color }) => {
   return (
-    <SvgIcon
-      htmlColor={color}
-      width="17"
-      height="17"
-      viewBox="0 0 17 17"
-      fill="none"
-    >
+    <SvgIcon htmlColor={color} width="17" height="17" viewBox="0 0 17 17">
       <path
         fillRule="evenodd"
         clipRule="evenodd"
--- a/components/events/EventsList.tsx
+++ b/components/events/EventsList.tsx
@@ -1,0 +1,152 @@
+import React, { useEffect, useState } from "react";
+
+import makeStyles from "@material-ui/core/styles/makeStyles";
+import { IconButton } from "@material-ui/core";
+import EventCardLarge from "./EventCardLarge";
+import EventCardGlimmerLarge from "./EventCardGlimmerLarge";
+import ChevronRightIcon from "@horizon/icons/ChevronRightIcon";
+import ChevronLeftIcon from "@horizon/icons/ChevronLeftIcon";
+
+const useStyles = makeStyles({
+  button: {
+    backgroundColor: "white",
+    boxShadow: "inset 0 0 0 1px #F0F0F0",
+    "&:hover": {
+      backgroundColor: "#f5f5f5"
+    }
+  },
+  nextButtonContainer: {
+    marginLeft: -24,
+    // boxShadow: "0px 5px 10px rgba(0, 0, 0, 0.05)",
+    borderRadius: "50%"
+  },
+  prevButtonContainer: {
+    marginLeft: 8,
+    marginRight: -56,
+    borderRadius: "50%",
+    position: "relative",
+    zIndex: 1
+  },
+  container: {
+    display: "flex",
+    flexDirection: "row",
+    alignItems: "center",
+    position: "relative",
+    marginLeft: -32
+  },
+  item: {
+    marginLeft: 32
+  }
+});
+
+interface EventDisplay {
+  name: string;
+  nonprofitName: string;
+  time: string;
+  imagePath: string;
+}
+
+const maxData = 14;
+let dataIndex = 0;
+const mockData = (i: number) => {
+  return {
+    name: `Feeding ${i + 1} Pigeon${i !== 0 ? "s" : ""} in the Park`,
+    nonprofitName: "Pigeon Feeders International",
+    time: "Sunday Morning",
+    imagePath: "defaultImages/defaultEvent.png"
+  };
+};
+
+const getEvents = async (): Promise<[EventDisplay[], boolean]> => {
+  // mock a server response lol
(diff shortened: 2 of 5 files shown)


components/core/icons/ChevronLeftIcon.tsx
```diff
@@ -0,0 +1,26 @@
+import { SvgIcon } from "@material-ui/core";
+import React from "react";
+
+interface Props {
+  color: string;
+}
+
+const ChevronLeftIcon: React.FC<Props> = ({ color }: Props) => {
```
PC98, line 8: nit: I forgot to mention this in your previous PRs, but `({ color }: Props)` can simply be written as `({ color }`, since we already tell React the prop types via `React.FC<Props>`

components/events/EventsList.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+const EventsList: React.FC = () => {
+  const classes = useStyles();
+
+  const [events, setEvents] = useState<EventDisplayInfo[]>([]);
+  const [numEvents, setNumEvents] = useState(0);
+  // Index of the first element displayed in the list
+  const [first, setFirst] = useState(0);
+  // Number of elements displayed
+  const [rowSize, setRowSize] = useState(4);
```
PC98, line 90: nit: declare 4 as a constant with a meaningful name in all caps outside this component

components/events/EventsList.tsx
```diff
@@ -0,0 +1,203 @@
...
+  }
+  return [out, dataIndex < maxData];
+};
+/** END MOCK SETUP */
+
+const EventsList: React.FC = () => {
+  const classes = useStyles();
+
+  const [events, setEvents] = useState<EventDisplayInfo[]>([]);
+  const [numEvents, setNumEvents] = useState(0);
```
PC98, line 86: Question: is `numEvents` just the length of `events`? If so, we can avoid creating state for it.

components/events/EventsList.tsx
```diff
@@ -0,0 +1,203 @@
...
+    display: "flex",
+    flexWrap: "wrap",
+    flexDirection: "row",
+    alignItems: "center",
+    position: "relative",
+    height: 270,
+    overflowX: "visible",
+    overflowY: "hidden",
+    marginLeft: -24,
+    paddingLeft: 24
```
PC98, line 40: out of curiosity, what's happening here? Is the net effect 0?

components/events/EventsList.tsx
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }, [numEvents, first, rowSize]);
+
+  const handleResize = useCallback(() => {
+    // wrap the resize in a 100ms debounce to prevent excess polling
+    clearTimeout(resizeTimeoutRef.current);
+    const w = containerRef.current?.offsetWidth;
+    resizeTimeoutRef.current = window.setTimeout(() => {
+      if (w != null) {
+        setRowSize(Math.max(1, Math.floor((w - 24) / 283)));
```
PC98, line 125: Let's try to avoid "magic" numbers.. declaring 24 and 283 as constants with meaningful names will be helpful.

components/events/EventsList.tsx
```diff
@@ -0,0 +1,203 @@
...
+  const [first, setFirst] = useState(0);
+  // Number of elements displayed
+  const [rowSize, setRowSize] = useState(4);
+  const [maxElem, setMaxElem] = useState(-1);
+  const [loading, setLoading] = useState(false);
+
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  // Lock calls to getEvents if one is currently in progress
+  const fetchingRef = useRef<boolean>(false);
```
PC98, line 97: nit: I think TS can infer this ref's type

components/events/EventsList.tsx
```diff
@@ -0,0 +1,203 @@
...
+    setFirst(Math.max(first - rowSize, 0));
+  };
+
+  // holds all the elements currently displayed - EventDisplayInfo gets rendered as
+  // EventCardLarge, while null gets rendered as EventCardGlimmerLarge
+  const display: (EventDisplayInfo | null)[] = events.slice(
+    first,
+    first + rowSize
+  );
+  if (maxElem == -1 || first + rowSize < maxElem) {
```
PC98, line 154: nit: `===`

components/events/EventsPage.tsx
```diff
@@ -17,7 +18,9 @@ const EventsPage: React.FC<Props> = props => {
         />
       }
     >
-      {null}
+      <div style={{ padding: 64, width: "100%" }}>
```
PC98, line 21: nit: ideally we will want to define these styles as we have been in other components (using material-ui).

components/core/icons/ChevronLeftIcon.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <SvgIcon
+      htmlColor={color}
+      width="17"
+      height="17"
+      viewBox="0 0 17 17"
+      fill="none"
```
PC98, line 15: Question: what's the difference between `fill` and `htmlColor`? 🤔

PC98 (approved):
🔥

components/events/EventsList.tsx
```diff
@@ -0,0 +1,203 @@
...
+    setFirst(first + rowSize);
+  };
+
+  const prevPage = () => {
+    setFirst(Math.max(first - rowSize, 0));
+  };
+
+  // holds all the elements currently displayed - EventDisplayInfo gets rendered as
+  // EventCardLarge, while null gets rendered as EventCardGlimmerLarge
+  const display: (EventDisplayInfo | null)[] = events.slice(
```
PC98, line 150: nit: let TS infer this

components/events/EventsList.tsx
migueldlr: Good catch, I think I had them separate for some reason earlier

components/events/EventsList.tsx
migueldlr: Yeah for some reason `overflow-y: hidden` doesn't play well with `overflow-x`:

image

components/core/icons/ChevronLeftIcon.tsx
migueldlr: Whoops we can take out `fill`

components/events/EventsList.tsx
migueldlr: We push `null`s into the array so we can't 😢

PR #209, opened 2020-09-29
Joynerjiang/170 navbar

savannahjoyner:
commented out the logo because it wasn't recognizing the png file. Let me know if you know a fix.


savannahjoyner:
image

components/core/HorizonButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+const useStyles = makeStyles({
+  root: {
+    borderRadius: 100,
+    boxShadow: "0px 5px 10px rgba(0, 0, 0, 0.05)",
+    background: "#FD8033",
+    color: "white"
+  }
+});
+
+const HorizonButton: React.FC<React.ComponentProps<typeof Button>> = ({
```
PC98, line 14: Doesn't a HorizonButton component exist already in develop? Can you try merging develop into your branch?

components/core/HorizonNavBar.tsx
```diff
@@ -0,0 +1,183 @@
...
+        <p>Events</p>
+      </MenuItem>
+      <MenuItem>
+        <p>Non-profits</p>
+      </MenuItem>
+      <MenuItem>
+        <p>Sign In</p>
+      </MenuItem>
+      <MenuItem>
+        <p>Sign Up</p>
```
PC98, line 126: Nit: to reduce some duplication with the code below, do you think we can make an array outside this component, `const NAV_TABS = ["Discover", "Events", ...]` and then `.map` through the array to construct the UI for mobile and desktop? This will help if let's say we decide to rename the "Discover" tab in the future, so the change will only have to be made in one place.

components/core/HorizonNavBar.tsx
```diff
@@ -0,0 +1,183 @@
...
+      </MenuItem>
+    </Menu>
+  );
+
+  return (
+    <div className={classes.grow}>
+      <AppBar position="static" style={{ background: "#FFFFFF" }}>
+        <Toolbar>
+          {/* <img src={Logo} alt="logo" className={classes.Logo} /> */}
```
PC98, line 135: Can you check how the `components/Branding.tsx` component does it, and try it that way?

PC98 (commented):
This looks good! Thanks for also making a mobile-friendly implementation! 💯

savannahjoyner:
image
new screen shot (spacing may have changed a bit)

PC98 (approved):

PR #218, opened 2020-09-30
Create card components for causes

### LINKED ISSUE #214: [VMS] Create card components for causes
* Similar to `EventCardLarge` and `EventCardGlimmerLarge` which contain UI specific to displaying Event details, create two new card components that will be used to display causes for the "Volunteer for a Cause" list.
* The cards should support an on-click event (this will be passed via props).
* Use the sample image in the Figma designs as the card background. You will have to download the image and put in somewhere in `public/`

matthebrown:
Closes #214


components/events/CauseCard.tsx
```diff
@@ -0,0 +1,73 @@
...
+  imagePath: string;
+  onClick: () => void;
+}
+
+const CauseCardLarge: React.FC<Props> = ({ cause, imagePath, onClick }) => {
+  const { card, cardContainer, textContainer, image, causeText } = useStyles();
+
+  return (
+    <div className={cardContainer}>
+      <div
```
matthewbrown77, line 57: This was what I found for adding `onClick` functionality to a general component like a div. If there is a better solution, please let me know. This was just what I found from a quick google search.

matthewbrown77:
cause_cards
Here is how the components currently look. They're essentially finished, but for some reason, I can't get the font quite right. I copied the exact CSS from the Figma [here](https://www.figma.com/file/1VBjx7n3zFQMng4YUNX6iz/Horizon-%2F-Web?node-id=418%3A58), but I still can't get the font to look right.

components/events/CauseCard.tsx
PC98: Thanks for flagging this! Can you try using the `FocusVisibleOnly` component that @migueldlr made? Or does that make the UI look different?

@migueldlr feel free to chime in here with thoughts

components/events/CauseCard.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+});
+
+interface Props {
+  cause: string;
+  imagePath: string;
+  onClick: () => void;
+}
+
+const CauseCardLarge: React.FC<Props> = ({ cause, imagePath, onClick }) => {
```
PC98, line 52: nit: let's try to prefix all component names with the folder they are in, so `Events*Whatever*`.

components/events/CauseCard.tsx
```diff
@@ -0,0 +1,73 @@
...
+    width: "inherit",
+    objectFit: "cover"
+  },
+  textContainer: {
+    position: "absolute",
+    width: 280,
+    height: 68,
+    left: "calc(50% - 280px/2)",
+    top: "calc(50% - 68px/2)"
+  },
```
PC98, line 33: This is great for now! Maybe in the future we should figure out a way to better share these styles with the glimmer component, or instead of having a different glimmer component, have one component that supports some sort of prop. Just something to think about 🤔

PC98 (approved):
Yeah, don't worry about Typography too much for now, I will look into it this weekend :)

components/events/CauseCard.tsx
migueldlr: Yup, this is definitely where I'd use `FocusVisibleOnly`

matthewbrown77: Did not know we had this component. Thanks for making this Miguel!

migueldlr:
Figma shows a card that's 246x138 while the current implementation has it as 358x200. Is there a reason for this? @PC98 @matthewbrown77 

Edit: I just realized the base component is 360x202, gonna reach out to design and ask what's up with it

PR #220, opened 2020-10-03
added api endpoint to call getUpcomingEventsCardData

schen2425:
added api endpoint to call getUpcomingEventsCardData

### DIFF
--- a/pages/api/getUpcomingEvents.ts
+++ b/pages/api/getUpcomingEvents.ts
@@ -5,6 +5,9 @@ import { handleRequestWithPayloadResponse } from "utils/util";
 // @route   GET event card data
 // @desc    Gets event data
 // @access  Public
+// even though isLastPage is not used by getUpcomingEventsCardData(),
+// the request body should include it since getUpcomingEventsCardData() 's
+// arguments have been typed to be of type DatePageWhatever
 export default async (
   req: NextApiRequest,
   res: NextApiResponse
@@ -12,5 +15,6 @@ export default async (
   handleRequestWithPayloadResponse(req, res, getUpcomingEventsCardData, [
     "date",
     "page",
-    "totalCount"
+    "totalCount",
+    "isLastPage"
   ]);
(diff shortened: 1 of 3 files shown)


requests/events.ts
```diff
@@ -0,0 +1,19 @@
+import config from "config";
+import { fetchRequestWithPayloadResponse } from "utils/util";
+import { DatePageInformation, DatePaginatedEventCards } from "utils/types";
+
+export const getUpcomingEvents = async (
+  date: string,
+  page: number,
+  totalCount: number
+): Promise<DatePageInformation> =>
```
PC98, line 9: `async ({date, page, totalCount, isLastPage}:  DatePageInformation)` so that we don't have to maintain the typings in two places

requests/events.ts
```diff
@@ -0,0 +1,19 @@
...
+  fetchRequestWithPayloadResponse<DatePaginatedEventCards>(
+    config.apis.getUpcomingEvents,
+    {
+      method: "get",
+      headers: {
+        "Content-Type": "application/json"
+      },
+      body: JSON.stringify({ date, page, totalCount })
```
PC98, line 17: nit: let's also pass `isLastPage`. `getUpcomingEventsCardData()` doesn't use it for now but it is part of the `DatePageInformation` type

pages/api/getUpcomingEvents.ts
```diff
@@ -0,0 +1,16 @@
...
+// @desc    Gets event data
+// @access  Public
+export default async (
+  req: NextApiRequest,
+  res: NextApiResponse
+): Promise<void> =>
+  handleRequestWithPayloadResponse(req, res, getUpcomingEventsCardData, [
+    "date",
+    "page",
+    "totalCount"
```
PC98, line 15: add `"isLastPage"` to this array as well.

requests/events.ts
PC98: **Edit:** I wonder if `async (datePageInformation: DatePageInformation)` will be enough; can you check if `JSON.stringify(datePageInformation)` will work as expected?

PC98 (approved):

PR #223, opened 2020-10-03
filter by cause

Fattimo:
created function in events.ts to filter by cause. imported said function into events/index.tsx for use when filter queries do exist. created common method in utils.util.ts to parse a query and return a corresponding array

### DIFF
--- a/components/events/useRouterQueryParamsState.tsx
+++ b/components/events/useRouterQueryParamsState.tsx
@@ -1,19 +1,12 @@
 import { useCallback, useMemo } from "react";
 import { useRouter } from "next/router";
+import { returnQueryAsArray } from "../../utils/util";
 
 const useRouterQueryParamsState = (key: string) => {
   const router = useRouter();
 
   const query = router.query[key];
-  const currentState = useMemo(() => {
-    if (query == null) {
-      return [];
-    }
-    if (Array.isArray(query)) {
-      return query;
-    }
-    return [query];
-  }, [query]);
+  const currentState = useMemo(() => returnQueryAsArray(query), [query]);
 
   const routerPushWithUpdatedState = useCallback(
     (updatedState: string[]) => {
--- a/components/events/useRouterQueryParamsState.tsx
+++ b/components/events/useRouterQueryParamsState.tsx
@@ -1,6 +1,6 @@
 import { useCallback, useMemo } from "react";
 import { useRouter } from "next/router";
-import { returnQueryAsArray } from "../../utils/util";
+import { returnQueryAsArray } from "utils/util";
 
 const useRouterQueryParamsState = (key: string) => {
   const router = useRouter();
--- a/pages/events/index.tsx
+++ b/pages/events/index.tsx
@@ -5,8 +5,8 @@ import {
   GetServerSidePropsContext
 } from "next";
 import EventsPage from "components/events/EventsPage";
-import { Dropdown } from "../../utils/types";
-import { returnQueryAsArray } from "../../utils/util";
+import { Dropdown } from "utils/types";
+import { returnQueryAsArray } from "utils/util";
 import { getCauses } from "server/actions/nonprofit";
 import {
   getUpcomingEventsCardData,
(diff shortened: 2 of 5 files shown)


Fattimo:
image
I was unable to test my function to filter by cause (in fact, I'm unsure if double `find()` is valid syntactically in the first place), as I kept on getting this error message regardless of how I passed `["OTHER"]` (I tried [OTHER], "OTHER", ["OTHER"], "[OTHER]")

server/actions/events.ts
```diff
@@ -112,6 +112,19 @@ export async function getNearestEventsCardDataCount({
   });
 }
 
+export async function getByCausesEventsCardData(causes: string[]) {
+  await Mongo();
+
+  const result = await Event.find()
+    .populate("nonprofitId", "cause", Nonprofit)
+    .find({
+      "nonprofitID.cause": { $elemMatch: { $in: causes } }
```
Fattimo, line 121: May be syntactical error and should be `nonprofitId`, but something to test

PC98:
@Fattimo Can you try testing using `'getByCausesEventsCardData(["OTHER"])'`; that doesn't throw an error for me:
```bash
$ yarn run:server:action events 'getByCausesEventsCardData(["OTHER"])'
yarn run v1.22.0
$ yarn run:ts-node:script server/run-action.ts events 'getByCausesEventsCardData(["OTHER"])'
$ cross-env NODE_ENV=development ts-node -O {\"module\":\"commonjs\"} -r module-alias/register server/run-action.ts events 'getByCausesEventsCardData(["OTHER"])'
[]
✨  Done in 12.18s.
```

components/events/useRouterQueryParamsState.tsx
```diff
@@ -1,19 +1,12 @@
 import { useCallback, useMemo } from "react";
 import { useRouter } from "next/router";
+import { returnQueryAsArray } from "../../utils/util";
```
PC98, line 3: nit: `from "utils/util";`

pages/events/index.tsx
```diff
@@ -6,10 +6,12 @@ import {
 } from "next";
 import EventsPage from "components/events/EventsPage";
 import { Dropdown } from "../../utils/types";
+import { returnQueryAsArray } from "../../utils/util";
```
PC98, line 9: nit: `from "utils/util";`
(also for some reason the above line also uses this :O, if you wanna fix that, that would be great).

pages/events/index.tsx
```diff
@@ -49,6 +51,9 @@ export const getServerSideProps = async (
       totalCount: upcomingEventsTotalCount,
       isLastPage: false
     });
+  } else {
+    const query = returnQueryAsArray(Object.keys(context.query));
```
PC98, line 55: This might not have the effect you are thinking 😅
`returnQueryAsArray(context.query["cause"])` might be more appropriate, since that's how the front-end populated the URL, `/events?time=WEEK`

server/actions/events.ts
Fattimo: it is `nonprofitId`

Fattimo:
image
Kinda something weird I noticed while testing my query; I'm not sure if this is a consistency thing/something that needs a migration script for the local db, or just my relative unfamiliarity with the structure of the Nonprofit schema. Essentially, the query only works when finding the field `causes` as opposed to just `cause`. While logging some of the outputs from Nonprofit, however, as seen below
image
there exists both a cause and a causes field. Is there a reason why the query only works when searching for the field `causes` and not `cause`?

PC98:
@Fattimo Oh okay there is an error in our migration script, it uses "causes": https://github.com/GTBitsOfGood/gen-soln/blob/develop/server/migrations/20200924180429-migrate-nonprofit-add-causes.ts#L7 whereas Mongoose schema uses "cause".

We can either update our Mongoose schema to use "causes" (it's okay if it still remains type String) OR write a new migration script that renames the "causes" field to "cause" for all Nonprofit documents.

server/actions/events.ts
```diff
@@ -112,6 +112,23 @@ export async function getNearestEventsCardDataCount({
...
 
+export async function getByCausesEventsCardData(causes: string[]) {
+  await Mongo();
+
+  const idsWithCause = await Nonprofit.find(
+    {
+      cause: { $in: causes }
+    },
+    "nonprofitId"
+  );
```
PC98, line 123: nit: I think you can use `distinct("_id", {cause: { $in: causes }})`

PC98 (approved):

PR #226, opened 2020-10-04
Kris/address autocomplete

KoolCards:
- Adds the autocomplete text field the contact step of the donation page
- Adds some optional parameters to the LocationAutocompleteInput component (fullWidth and defaultValue), but I think VMS may need these parameters as well, so they can be converted to required fields whenever VMS wants to update their references to this component

### DIFF
--- a/components/donation/form/DonationPageFormContactStep.tsx
+++ b/components/donation/form/DonationPageFormContactStep.tsx
@@ -52,7 +52,10 @@ const DonationPageFormContactStep: React.FC<ContactStepProps> = ({
   } = useStyles();
   const dispatch = useContext(DonationPageStateDispatch);
 
-  const isCurStepCompleted = useMemo(() => address !== null, [address]);
+  const isCurStepCompleted = useMemo(
+    () => (firstName && lastName && email && address ? true : false),
+    [firstName, lastName, email, address]
+  );
 
   useEffect(() => {
     // For this step, we will defter to the browser's API to indicate when the required TextFields have been filled
--- a/components/donation/form/DonationPageFormContactStep.tsx
+++ b/components/donation/form/DonationPageFormContactStep.tsx
@@ -1,14 +1,18 @@
-import React, { useContext, useCallback, useEffect } from "react";
+import React, { useContext, useCallback, useEffect, useMemo } from "react";
 import clsx from "clsx";
 import makeStyles from "@material-ui/core/styles/makeStyles";
-
 import TextField from "@material-ui/core/TextField";
+
+import LocationAutocompleteInput, {
+  PlaceType
+} from "components/LocationAutocompleteInput";
 
 import {
   ContactStepProps,
   DonationPageStateDispatch,
   setContactStepField,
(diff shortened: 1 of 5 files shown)


components/donation/form/DonationPageFormContactStep.tsx
```diff
@@ -47,21 +52,33 @@ const DonationPageFormContactStep: React.FC<ContactStepProps> = ({
   } = useStyles();
   const dispatch = useContext(DonationPageStateDispatch);
 
+  const isCurStepCompleted = useMemo(
+    () => (firstName && lastName && email && address ? true : false),
+    [firstName, lastName, email, address]
+  );
+
```
PC98, line 59: Instead of doing this, can we make the `LocationAutocompleteInput` `required`? That's how this form component ensures that users have entered their names and email before proceeding.

components/donation/form/reducer.ts
```diff
@@ -11,6 +13,7 @@ export interface ContactStepProps {
   firstName: string;
   lastName: string;
   email: string;
+  address: PlaceType | null;
```
PC98, line 16: Definitely something to think about -- what are the pros and cons of storing `address` as a `PlaceType` as opposed to a simple `string`. This will depend on what we intend on using `address` for.

components/donation/form/DonationPageFormContactStep.tsx
```diff
@@ -99,6 +116,17 @@ const DonationPageFormContactStep: React.FC<ContactStepProps> = ({
         }}
         autoComplete="email"
       />
+      <div className={clsx(name, verticalPositiveMargin)}>
+        <LocationAutocompleteInput
+          addPlaceChip={address => {
+            onAddressChange(address);
+          }}
```
PC98, line 123: nit: can just do `addPlaceChip= {onAddressChange}`

components/donation/form/DonationPageFormContactStep.tsx
```diff
@@ -99,6 +116,17 @@ const DonationPageFormContactStep: React.FC<ContactStepProps> = ({
...
+          locationType="address"
+          fullWidth={true}
```
PC98, line 125: nit: just `fullWidth` as opposed to `fullWidth={true}` should work as well.

components/donation/form/reducer.ts
KoolCards: yep, so we actually need it to be a placetype because when u go from the contact tab to the amount tab and then back to the donate tab, we want the locationautocompleteinput component to still retain the placetype input value. Also, we use two attributes from the placetype object (main_text and secondary_text) so it seemed easier to just keep a placetype object that we can parse later.

components/donation/form/DonationPageFormContactStep.tsx
KoolCards: good point, fixed

parthnagpal17 (approved):

PC98 (approved):
LGTM (if you can fix the lint warnings, that would be great!)

PR #227, opened 2020-10-04
Implement upcoming volunteer events list

### LINKED ISSUE #212: [VMS] Implement "Upcoming Volunteer Events" list
* Modify the `EventsList` component **(note this is just a suggestion, feel free to think of your own ways to do this task!):** 
  - [x] It should take an object, `paginatedEventCardsData` of type `PaginatedEventCards` as props along with a function, something like `getMoreEvents`. `getMoreEvents` will take an integer, `newPage`as input and return an object of type `PaginatedEventCards`.
  - [x] Chage the `events` `useState` to be of type `EventCardData` (you can get this from `utils/types.ts`) and set the inital value to `paginatedEventCardsData.eventCards`.
  - [x] Create a new `useRef`, `pageRef` - set it's initial value to `paginatedEventCardsData.page`.
  - [x] Modify this line: `const [newEvents, hasMore] = await getEvents(fetchNum);` to: `const { newEvents : eventCards, newPage : page, isLastPage}  = await getMoreEvents(pageRef.current + 1)`. 
  - [x] Set `pageRef.current` to `newPage.`
  - [x] You will have to modify some other existing code to use `isLastPage` instead of `hasMore` and the `EventCardLarge` components will have to be rendered using the `EventCardData` type, instead of the `EventDisplayInfo` type created in this file. 
  - [ ] **Bonus points** if you can refactor `EventCardLarge`'s prop types to directly work with `EventCardData` instead of `headerText`, `bodyText`, etc.). Though let's talk to Miguel about this if he thinks this is a good idea. 
- [x] Then, make a component, something like `EventsPageMainContent`. This component will be rendered in `EventsPage.tsx` and for now, it should accept an object of type `DatePaginatedEventCards` as props (this was added to `EventsPage.tsx` as part of #211).
- [x] `EventsPageMainContent` should render the title text - "Upcoming Volunteer Events" and an `EventsList` component. Think about what props you want to pass to `EventsList`, particularly the `getMoreEvents` function. As a hint, `getMoreEvents` will use the fetch request added as part of #210. The input to the fetch request - `date` and `totalCount` it can get from its own props but the new `page` value be passed by `EventsList`.
- [ ] As an experiment, try adjusting the `CARDS_PER_PAGE` value defined in `server/events.ts` to see if that leads to a better UI overall.

matthewbrown77:
Closes #212

### DIFF
--- a/components/events/EventsList.tsx
+++ b/components/events/EventsList.tsx
@@ -6,7 +6,7 @@ import ChevronLeftIcon from "@horizon/icons/ChevronLeftIcon";
 
 import EventCardLarge from "./EventCardLarge";
 import EventCardGlimmerLarge from "./EventCardGlimmerLarge";
-import { PaginatedEventCards, EventCardData } from "../../utils/types";
+import { PaginatedEventCards, EventCardData } from "utils/types";
 
 const useStyles = makeStyles({
   button: {
--- a/components/events/EventsList.tsx
+++ b/components/events/EventsList.tsx
@@ -1,11 +1,12 @@
 import React, { useEffect, useState, useRef, useCallback } from "react";
-
 import makeStyles from "@material-ui/core/styles/makeStyles";
 import { IconButton } from "@material-ui/core";
-import EventCardLarge from "./EventCardLarge";
-import EventCardGlimmerLarge from "./EventCardGlimmerLarge";
 import ChevronRightIcon from "@horizon/icons/ChevronRightIcon";
 import ChevronLeftIcon from "@horizon/icons/ChevronLeftIcon";
+
+import EventCardLarge from "./EventCardLarge";
+import EventCardGlimmerLarge from "./EventCardGlimmerLarge";
+import { PaginatedEventCards, EventCardData } from "../../utils/types";
 
 const useStyles = makeStyles({
   button: {
@@ -44,55 +45,33 @@ const useStyles = makeStyles({
   }
 });
 
-interface EventDisplayInfo {
-  name: string;
-  nonprofitName: string;
-  time: string;
-  imagePath: string;
+interface Props {
+  paginatedEventCardsData: PaginatedEventCards;
+  getMoreEvents: (newPage: number) => Promise<PaginatedEventCards>;
 }
-
-/** BEGIN MOCK SETUP */
-const maxData = 14;
-let dataIndex = 0;
-const mockData = (i: number) => {
-  return {
-    name: `Feeding ${i} Pigeon${i !== 1 ? "s" : ""} in the Park`,
-    nonprofitName: "Pigeon Feeders International",
-    time: "Sunday Morning",
-    imagePath: "defaultImages/defaultEvent.png"
-  };
-};
-
-/**
- * Mocks a server request
- * @param n the number of events to return
- * @return a promise that contains [an array of EventDisplayInfo, true if there are remaining events]
- */
-const getEvents = async (n: number): Promise<[EventDisplayInfo[], boolean]> => {
-  await new Promise(r => setTimeout(r, 1000));
-  const out = [];
-  for (let i = 0; i < n && dataIndex < maxData; i++) {
-    out.push(mockData(dataIndex));
-    dataIndex++;
-  }
-  return [out, dataIndex < maxData];
-};
-/** END MOCK SETUP */
 
 const DEFAULT_ROW_SIZE = 4;
 
-const EventsList: React.FC = () => {
+const EventsList: React.FC<Props> = ({
+  paginatedEventCardsData,
+  getMoreEvents
+}) => {
   const classes = useStyles();
 
-  const [events, setEvents] = useState<EventDisplayInfo[]>([]);
+  const [events, setEvents] = useState<EventCardData[]>(
+    paginatedEventCardsData.eventCards
+  );
+
   // Index of the first element displayed in the list
   const [first, setFirst] = useState(0);
+
   // Number of elements displayed
   const [rowSize, setRowSize] = useState(DEFAULT_ROW_SIZE);
   const [maxElem, setMaxElem] = useState(-1);
   const [loading, setLoading] = useState(false);
 
   const containerRef = useRef<HTMLDivElement>(null);
+  const pageRef = useRef<number>(paginatedEventCardsData.page);
 
   // Lock calls to getEvents if one is currently in progress
   const fetchingRef = useRef(false);
@@ -104,11 +83,19 @@ const EventsList: React.FC = () => {
       void (async () => {
         fetchingRef.current = true;
         setLoading(true);
-        const fetchNum = first + rowSize - events.length;
-        const [newEvents, hasMore] = await getEvents(fetchNum);
-        if (!hasMore) {
+
+        const {
+          eventCards: newEvents,
+          page: newPage,
+          isLastPage
+        } = await getMoreEvents(pageRef.current + 1);
+
+        pageRef.current = newPage;
+
+        if (isLastPage) {
           setMaxElem(events.length + newEvents.length);
         }
+
         setEvents(prevEvents => [...prevEvents, ...newEvents]);
         setLoading(false);
         fetchingRef.current = false;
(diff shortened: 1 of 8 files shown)


components/events/EventsList.tsx
```diff
@@ -180,9 +167,9 @@ const EventsList: React.FC = () => {
           {event != null ? (
             <EventCardLarge
               headerText={event.name}
-              bodyText={event.nonprofitName}
-              metaText={event.time}
-              imagePath={event.imagePath}
+              bodyText={event.nonprofitId.name}
```
matthewbrown77, line 170: For some reason, I receive an undefined error when trying to access the TS defined field `nonprofitID` here. Maybe I'm misunderstanding how `Pick` works, but this worked for me.

components/events/EventsList.tsx
```diff
@@ -180,9 +167,9 @@ const EventsList: React.FC = () => {
...
+              metaText={event.duration.toString()} //TODO: replace with formatting from figma
+              imagePath={event.image}
```
matthewbrown77, line 172: Right now, the image is returning undefined as the application can't find the default image being set by the event model (`'url("/defaultImages/defaultEvent.png")'`). Can you explain why we are setting it this way?

requests/events.ts
```diff
@@ -2,16 +2,19 @@ import config from "config";
...
   fetchRequestWithPayloadResponse<DatePaginatedEventCards>(
     config.apis.getUpcomingEvents,
     {
-      method: "get",
+      method: "post",
       headers: {
         "Content-Type": "application/json"
       },
-      body: JSON.stringify({ datePageInformation })
+      body: JSON.stringify({ page, date, totalCount, isLastPage })
```
matthewbrown77, line 18: I had to make some changes to get the request to work properly. Firstly, since we are passing the parameters in the body we have to use post. Second, I used destructuring on the parameters to work better with the client.

matthewbrown77:
I'm still open to discussing any potential changes to `EventCardLarge` with @migueldlr, but the only advantage I see is that it improves `readability` (and there is probably a reason he decided to implement it how he did).

matthewbrown77:
event_list

components/events/EventsList.tsx
PC98: I think the problem we made a typo when defining the `EventCardData` type: https://github.com/GTBitsOfGood/gen-soln/blob/develop/utils/types.ts#L51 We should be using `nonprofitId` and not `nonprofitID`, since that's what Mongoose will return. This will fix your undefined error as well.

components/events/EventsList.tsx
PC98: Ah okay. If we were using CSS to set the image, then we would need the `url` but, but since `EventCardLarge` is using the `img` tag's `src` property, our DB should just be storing it as `"/defaultImages/defaultEvent.png"`.

Here is what you can do- modify `DEFAULT_IMAGE` in `event.js` file to be `"/defaultImages/defaultEvent.png"`. Create a migration script, called something like `seed-events-fix-image.ts` and the code in there can be exactly the same as the existing `20200925065734-seed-events-add-image.ts` script.

requests/events.ts
PC98: Does `JSON.stringify(datePageInformation)` work as expected?

PC98:
> I'm still open to discussing any potential changes to `EventCardLarge`

So the pro of changing `EventCardLarge.tsx`'s prop types to directly take an object of type`EventCardData` is that we encapsulate the details of what an event card's header, body, and meta text should be in the component itself.

The current approach of supplying `headerText`, `bodyText`, etc. as props to `EventCardLarge` would be appropriate if this same card UI was used in multiple places. Then we would even change the name of the component to something like `CoreCardLarge.tsx` (not specific to events).

components/events/EventsList.tsx
```diff
@@ -1,12 +1,13 @@
...
 import makeStyles from "@material-ui/core/styles/makeStyles";
 import { IconButton } from "@material-ui/core";
-import EventCardLarge from "./EventCardLarge";
-import EventCardGlimmerLarge from "./EventCardGlimmerLarge";
 import ChevronRightIcon from "@horizon/icons/ChevronRightIcon";
 import ChevronLeftIcon from "@horizon/icons/ChevronLeftIcon";
 
+import EventCardLarge from "./EventCardLarge";
+import EventCardGlimmerLarge from "./EventCardGlimmerLarge";
+import { PaginatedEventCards, EventCardData } from "../../utils/types";
```
PC98, line 9: nit: use the alias `from  "utils/types";`

components/events/EventsList.tsx
```diff
@@ -44,55 +45,33 @@ const useStyles = makeStyles({
...
   // Index of the first element displayed in the list
   const [first, setFirst] = useState(0);
+
   // Number of elements displayed
   const [rowSize, setRowSize] = useState(DEFAULT_ROW_SIZE);
   const [maxElem, setMaxElem] = useState(-1);
   const [loading, setLoading] = useState(false);
 
   const containerRef = useRef<HTMLDivElement>(null);
+  const pageRef = useRef<number>(paginatedEventCardsData.page);
```
PC98, line 74: nit: does TS not infer the `number` type for you? If so, you can drop it.

components/events/EventsPage.tsx
```diff
@@ -30,9 +21,9 @@ const EventsPage: React.FC<Props> = props => {
         />
       }
     >
-      <div className={classes.container}>
-        <EventsList />
-      </div>
+      <EventsPageMainContent
+        upcomingEvents={props.upcomingEventsFirstPageData!}
```
PC98, line 25: We will probably render a different component if `props.upcomingEventsFirstPageData` is undefined. For now, let's render `null`:

```javascript
{props.upcomingEventsFirstPageData && <EventsPageMainContent upcomingEvents={props.upcomingEventsFirstPageData} />}
```
If that doesn't work, you might have to destructure `props` on line 14:
`const EventsPage: React.FC<Props> = ({upcomingEventsFirstPageData, /* other stuff */ } ) => {`

components/events/EventsPageMainContent.tsx
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+import { Typography, makeStyles } from "@material-ui/core";
+
+import EventsList from "./EventsList";
+import { getUpcomingEvents } from "../../requests/events";
```
PC98, line 5: nit: use alias  `from "requests/events";`

requests/events.ts
```diff
@@ -2,16 +2,19 @@ import config from "config";
...
+  page,
+  date,
+  totalCount,
+  isLastPage
+}: DatePageInformation) =>
   fetchRequestWithPayloadResponse<DatePaginatedEventCards>(
     config.apis.getUpcomingEvents,
     {
-      method: "get",
+      method: "post",
```
PC98, line 14: thanks for catching this! I think @migueldlr will work on making this a GET request soon.

components/events/EventsPageMainContent.tsx
```diff
@@ -0,0 +1,56 @@
...
+const EventsPageMainContent: React.FC<Props> = ({ upcomingEvents }) => {
+  const { mainContainer, title, listContainer } = useStyles();
+
+  return (
+    <div className={mainContainer}>
+      <Typography className={title}>Upcoming Volunteer Events</Typography>
+      <div className={listContainer}>
+        <EventsList
+          paginatedEventCardsData={upcomingEvents}
+          getMoreEvents={(page: number) => {
```
PC98, line 42: nit: as an FYI, you can skip the parenthesis if returning directly in an arrow function:
```javascript
(page: number) => getUpcomingEvents({
               page,
               date: upcomingEvents.date,
               totalCount: upcomingEvents.totalCount,
               isLastPage: upcomingEvents.isLastPage
             })
```

components/events/EventsList.tsx
matthewbrown77: TS requires an argument here.

requests/events.ts
matthewbrown77: This did work as expected, so I reverted the destructure and removed the extra braces.

components/events/EventsList.tsx
matthewbrown77: Added in the latest commit.

components/events/EventsList.tsx
matthewbrown77: Fixed in latest commit.

PC98 (approved):
💯

PR #230, opened 2020-10-06
Add API endpoint and fetch request method, update utils

### LINKED ISSUE #213: [VMS] Create API endpoint and a fetch request method to call `getNearestEventsCardData()`
* Similar to #210, but for `getNearestEventsCardData`

migueldlr:
Closes #213 
Also modifies both `handleRequestWithPayloadResponse` and `fetchRequestWithPayloadResponse` in `utils.ts` to handle `GET` requests

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "next": "9.5.2",
     "next-auth": "^3.1.0",
     "next-cookies": "^2.0.3",
+    "querystringify": "^2.2.0",
     "react": "16.13.1",
     "react-dom": "16.13.1",
     "stripe": "^8.40.0"
@@ -55,6 +56,7 @@
     "@types/mongoose": "^5.7.36",
     "@types/next-auth": "^3.1.10",
     "@types/node": "^14.6.0",
+    "@types/querystringify": "^2.0.0",
     "@types/react": "^16.9.47",
     "@typescript-eslint/eslint-plugin": "^3.10.1",
     "@typescript-eslint/parser": "^3.10.1",
--- a/pages/api/getNearestEvents.ts
+++ b/pages/api/getNearestEvents.ts
@@ -1,0 +1,17 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import { getNearestEventsCardData } from "server/actions/events";
+import { handleRequestWithPayloadResponse } from "utils/util";
+
+// @route   GET nearest events
+// @desc    Gets nearest events
+// @access  Public
+export default async (
+  req: NextApiRequest,
+  res: NextApiResponse
+): Promise<void> =>
+  handleRequestWithPayloadResponse(req, res, getNearestEventsCardData, [
+    "date",
+    "page",
+    "totalCount",
+    "isLastPage"
+  ]);
--- a/pages/api/getNearestEvents.ts
+++ b/pages/api/getNearestEvents.ts
@@ -10,7 +10,7 @@ export default async (
   res: NextApiResponse
 ): Promise<void> =>
   handleRequestWithPayloadResponse(req, res, getNearestEventsCardData, [
-    "date",
+    "location",
     "page",
     "totalCount",
     "isLastPage"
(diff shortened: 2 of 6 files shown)


pages/api/getNearestEvents.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// @route   GET nearest events
+// @desc    Gets nearest events
+// @access  Public
+export default async (
+  req: NextApiRequest,
+  res: NextApiResponse
+): Promise<void> =>
+  handleRequestWithPayloadResponse(req, res, getNearestEventsCardData, [
+    "date",
```
PC98, line 13: Actually we want these args to be of type `LocationPageInformation`, so use `location` instead of `date`

requests/events.ts
```diff
@@ -1,17 +1,35 @@
...
     config.apis.getUpcomingEvents,
     {
-      method: "post",
+      method: "get",
       headers: {
         "Content-Type": "application/json"
-      },
-      body: JSON.stringify(datePageInformation)
-    }
+      }
```
PC98, line 18: nit: do you think we can skip `"Content-Type"` (here and below) now that we making GET requests?

utils/util.ts
```diff
@@ -45,13 +51,22 @@ export const handleRequestWithPayloadResponse = async <T>(
 
 // Use this function on client side to make API requests
 export const fetchRequestWithPayloadResponse = async <T>(
-  url: RequestInfo,
-  options: RequestInit = {}
+  url: string,
+  options: RequestInit = {},
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  body: object = {}
```
PC98, line 57: I know you said `Record<string, unkown>` doesn't work, but does `{[key in string]: unknown}` work? Will this let us get rid of the eslint-disable-next-line?

utils/util.ts
```diff
@@ -45,13 +51,22 @@ export const handleRequestWithPayloadResponse = async <T>(
...
 ): Promise<T> => {
-  const res = await fetch(url, {
+  const isGetRequest = options.method === "GET";
+
+  const fullUrl = isGetRequest ? `${url}&${qs.stringify(body)}` : url;
```
PC98, line 61: Should it be `${url}?${qs.stringify(body)}` since the "?" character separates the URL path from query parameters?

utils/util.ts
```diff
@@ -45,13 +51,22 @@ export const handleRequestWithPayloadResponse = async <T>(
...
+
+  const fullOptions: RequestInit = {
     mode: "same-origin",
-    ...options
-  });
+    ...options,
+    body: isGetRequest ? undefined : JSON.stringify(body)
+  };
```
PC98, line 67: perhaps just inline it with the fetch call as it was before as opposed to creating a variable and typing it to `RequestInit` (unless TS is making you do it this way).

package.json
```diff
@@ -37,6 +37,7 @@
     "next": "9.5.2",
     "next-auth": "^3.1.0",
     "next-cookies": "^2.0.3",
+    "querystringify": "^2.2.0",
```
PC98, line 40: Question: what are some pros/cons of using `URLSearchParams`: https://stackoverflow.com/questions/35038857/setting-query-string-using-fetch-get-request as opposed to a package?

requests/events.ts
```diff
@@ -1,17 +1,35 @@
...
+    },
+    datePageInformation
+  );
+
+export const getNearestEvents = async (
+  locationPageInformation: LocationPageInformation
+) =>
+  fetchRequestWithPayloadResponse<DatePaginatedEventCards>(
```
PC98, line 26: use `LocationPaginatedEventCards`

utils/util.ts
migueldlr: It lets us remove the ESLint disable, but we end up with a TS error when we try to pass in an object of type `DatePageInformation` or `LocationPageInformation` at the callsite. (`Index signature is missing in type`, specifically)

package.json
migueldlr: Pros:
- Smaller (though we could improve the current code by only importing the `stringify` method rather than the entire package)
- Reduces dependencies
- We could _maybe_ have finer control over how args are set in the query (though we really shouldn't need this?)

Cons:
- No IE support
- No pretty one-liner ternary statement (the constructor `URLSearchParams` wants `Record<string, string>` so we're forced to loop over individual keys)

PC98 (approved):

PR #231, opened 2020-10-06
UI for Signup and Login

kw143:
This pull request adds a barebone home page with a message and two buttons. The message states whether the user is logged in based on session status returned by Auth0 service. Home page can be found at "/home".
image

This pull request adds basic Next-Auth configuration file.
This pull request adds a sign in page provided by Next-Auth with various sign in options. This page can be found by clicking on the sign in button at home page.
image
This sign-in page will not work unless if an Auth-0 account is configured locally(by modifying .env file with Auth0 credentials).

This pull request adds a sign up page created from scratch based on Figma. The sign up button on home page leads to this page.
image
The styling could not be made exact and some of the colors was overwritten locally due to mismatch in typography, Horizon Link and Figma design: The custom Typography in the project currently does not match with Figma, specfically the "h3" variance, and font colors of all variances. The horizon link also does not appear to have the correct orange color.


rithikgavvala (approved):
Good work Kevin,  I will be merging in the in the Typography but this is close enough.

PR #248, opened 2020-10-08
Call getNearestEventsCardDataCount in getNearestEventsCardData

### LINKED ISSUE #241: [VMS] Call `getNearestEventsCardDataCount` in `getNearestEventsCardData`
* Unlike `getUpcomingEventsCardData()` which first gets called on the server-side and receives the result of `getUpcomingEventsCardDataCount()`, `getNearestEventsCardDataCount()` cannot be called server-side since a user's location is not known.
* This means that `getNearestEventsCardData()` is first called by the client via an API call with some invalid/dummy value for `totalCount` (perhaps -1). You should modify the code in `getNearestEventsCardData` to check if `totalCount` equals this dummy value and call `getUpcomingEventsCardDataCount` if that's the case.
* For clean-cpde purposes, let's also modify `getNearestEventsCardDataCount` arguments to be of type `LocationPageInformation["location"]`. Then you can stop exporting `Coordinates` from `utils/types.ts`

matthewbrown77:
Closes #241

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -4,7 +4,6 @@ import Nonprofit from "server/models/nonprofit";
 import errors from "utils/errors";
 
 import {
-  Coordinates,
   Event as EventType,
   EventCardData as EventCardDataType,
   DatePaginatedEventCards,
@@ -74,6 +73,10 @@ export async function getNearestEventsCardData({
 }: LocationPageInformation): Promise<LocationPaginatedEventCards> {
   await Mongo();
 
+  if (totalCount == -1) {
+    totalCount = await getNearestEventsCardDataCount(location);
+  }
+
   const result = await Event.find(
     {
       "address.location": {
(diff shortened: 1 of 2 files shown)


matthewbrown77:
Output from calling `getNearestEventsCardData` with a count of `-1`:
result

server/actions/events.ts
```diff
@@ -74,6 +73,10 @@ export async function getNearestEventsCardData({
 }: LocationPageInformation): Promise<LocationPaginatedEventCards> {
   await Mongo();
 
+  if (totalCount == -1) {
```
PC98, line 76: for future reference, consider using strict equality `===` 🙂

PC98 (approved):

server/actions/events.ts
matthewbrown77: Is using strict equality necessary with typescript?

PC98: Not really, but I guess this is a general rule of thumb when working with a mixed codebase (JS and TS) and a language that supports both strict and non-strict equality operators.

Issue #242, opened 2020-10-07
[VMS] Implement the "Volunteer Events Near You" list


PC98:
* For inspiration, look at how `EventsPageMainContent` renders the "Upcoming Volunteer Events" list.
* You will have to use HTML 5's Geocoding API to get the user's latitude and longitude information (potentially in a `useEffect` hook) and then make a call to `getNearestEventsCardData` via the request added as part of #213.
* The initial value for `isLastPage` can be false, `page` will be 0 and `totalCount` will be whatever invalid value is considered in #241.
* If a user refuses to share their location, or if the backend doesn't return any events near their location, we should not render this list component at all.

matthewbrown77:
@PC98 Is there a better way to address conditional rendering rather than just having to check each prop field? For example, with just `upcomingEventsFirstPageData` and `nearestEventsFirstPageData`, we need to have 3 conditional renders checking each possible combination, which leads to a lot of duplicated code.

matthewbrown77:
Actually, I think we could just move the conditional render into the `MainContent` component instead of `EventsPage`, which is where it currently is.

PC98:
So actually when I said:

>  if the backend doesn't return any events near their location, we should not render this list component at all.

I meant that `EventsPageMainContent` should call the request added as part of #213 after it renders. Once we have the result from the request, it should render a`EventsPageEventList` component corresponding to the nearest events if the server returned a non-empty array.

I believe you are talking about the `props.upcomingEventsFirstPageData &&` check in `EventsPage.tsx`. I don't think you will have to make any changes to that as part of this issue.
Though I agree with the fact that checking `props.upcomingEventsFirstPageData` to determine if `EventsPageMainContent` should be rendered hides the fact that `EventsPageMainContent` is actually responsible for rendering other lists of data as well. I can put out a PR to make this more explicit.

matthewbrown77:
@PC98 GET requests are still broken on the `develop` branch. If you visit our [vercel hosted build](bog-gen-soln.vercel.app) and navigate to `/events`, you can see that the initial upcoming events are never loaded, and that adding and then removing a filter sends a request that returns a 404.
![404](https://user-images.githubusercontent.com/45107667/95480929-73d55700-095a-11eb-9999-a0800bb31b4f.png)

PC98:
Thanks for flagging, I will look into this

matthewbrown77:
[...]

PR #249, opened 2020-10-08
Create a different component to handle /events page with and without query

PC98:
As a follow-up to this: https://github.com/GTBitsOfGood/gen-soln/issues/242#issuecomment-705329233, the `EventsPage.tsx` component was really functioning as two different components. One that renders the various lists (like events near you) when no query parameters are specified and one that _(eventually will)_ render events in a grid-like manner that satisfy the passed in query parameters.

Making a different component and using some TS types to make this more explicit.


pages/events/index.tsx
```diff
@@ -40,27 +57,37 @@ export const getServerSideProps = async (
...
       page: 0,
       totalCount: upcomingEventsTotalCount,
       isLastPage: false
     });
+
+    return {
+      props: {
+        ...commonProps,
+        upcomingEventsFirstPageData,
+        type: "WITHOUT_QUERY" as const
```
PC98, line 79: @migueldlr @matthewbrown77 might be of interest to you: this is how we can define discriminated union types in TypeScript! A very powerful tool that you should lookup. 

`LocationAutocompleteInput.tsx` also uses something similar when defining its prop types.

matthewbrown77: Very helpful! Thanks for the pro-tip 🚀

matthewbrown77 (approved):

matthewbrown77 (approved):

PR #251, opened 2020-10-08
Make checking for GET in fetchRequestWithPayloadResponse case-insensitive

### LINKED ISSUE #250: GET requests are broken for events requests
#230 tried to implement GET requests in `fetchRequestWithPayloadResponse`, but there's faulty logic.

We use `"get"` at the callsite, but checks for `"GET"` in the function so it always sends a POST.

migueldlr:
Closes #250

### DIFF
--- a/utils/util.ts
+++ b/utils/util.ts
@@ -20,7 +20,7 @@ export const handleRequestWithPayloadResponse = async <T>(
   callback: (body: NextApiRequest["body"]) => Promise<T>,
   bodyHasProperties: string[] = []
 ): Promise<void> => {
-  const isGetRequest = req.method === "GET";
+  const isGetRequest = req.method?.toUpperCase() === "GET";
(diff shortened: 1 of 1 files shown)


utils/util.ts
```diff
@@ -56,7 +56,7 @@ export const fetchRequestWithPayloadResponse = async <T>(
   // eslint-disable-next-line @typescript-eslint/ban-types
   body: object = {}
 ): Promise<T> => {
-  const isGetRequest = options.method === "GET";
+  const isGetRequest = options.method?.toUpperCase() === "GET";
```
PC98, line 59: Do we want to make a similar change on line 23, or is that always correct set by Next.js?

PC98 (approved):

utils/util.ts
migueldlr: Oop

PR #257, opened 2020-10-08
refactor for EventCardData in props for EventCards

Fattimo:
refactor with type eventCardData. moved about to be apart of type Eve…ntBase so that there is body text for the eventCard to access

### DIFF
--- a/components/events/EventsPageEventCard.tsx
+++ b/components/events/EventsPageEventCard.tsx
@@ -97,7 +97,7 @@ const EventsPageEventCard: React.FC<Props> = ({ eventCardData, onClick }) => {
               {eventCardData.name}
             </CoreTypography>
             <CoreTypography className={clsx(body, truncate)}>
-              {eventCardData.name}
+              {eventCardData.nonprofitId.name}
             </CoreTypography>
           </div>
         </div>
--- a/components/events/EventsPageEventCard.tsx
+++ b/components/events/EventsPageEventCard.tsx
@@ -97,7 +97,7 @@ const EventsPageEventCard: React.FC<Props> = ({ eventCardData, onClick }) => {
               {eventCardData.name}
             </CoreTypography>
             <CoreTypography className={clsx(body, truncate)}>
-              {eventCardData.name /*TODO: find body for event card */}
+              {eventCardData.name}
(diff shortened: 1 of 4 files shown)


components/events/EventsPageEventCard.tsx
```diff
@@ -92,16 +82,22 @@ const EventsPageEventCard: React.FC<Props> = ({
...
+                eventCardData.duration /*TODO: replace with formatting from figma*/
+              }
             </CoreTypography>
             <CoreTypography variant="h4" className={header}>
-              {headerText}
+              {eventCardData.name}
             </CoreTypography>
             <CoreTypography className={clsx(body, truncate)}>
-              {bodyText}
+              {eventCardData.about}
```
PC98, line 100: actually we want this to be the nonprofit name, so `eventCardData.nonprofitId.name`

server/actions/events.ts
```diff
@@ -19,7 +19,8 @@ const CARD_FIELDS: Record<keyof EventCardDataType, 1> = {
   image: 1,
   address: 1,
   nonprofitId: 1,
-  duration: 1
+  duration: 1,
+  about: 1
```
PC98, line 23: you should undo this change, and the changes to `EventBase` and `Event` types

components/events/EventsPageEventCard.tsx
```diff
@@ -92,16 +82,22 @@ const EventsPageEventCard: React.FC<Props> = ({
...
+                eventCardData.duration /*TODO: replace with formatting from figma*/
+              }
             </CoreTypography>
             <CoreTypography variant="h4" className={header}>
-              {headerText}
+              {eventCardData.name}
             </CoreTypography>
             <CoreTypography className={clsx(body, truncate)}>
-              {bodyText}
+              {eventCardData.name}
```
PC98, line 100: Should be `eventCardData.nonprofitId.name` 🙂

Fattimo: Wow that was really bad from me, I didn't read the original message closely enough 😅

PC98: np!

PC98 (approved):

PR #260, opened 2020-10-09
Removes 'Upcoming Volunteer Events' when no upcoming events

### LINKED ISSUE #244: [VMS] Don't render the "Upcoming Volunteer Events" text when there are none
* Right now, in `EventsPageMainContent`, we render the header "Upcoming Volunteer Events" even when `upcomingEvents.eventCards` is an empty array. 
* However we should render the text nor the `EventsPageEventList` if that is the case.

jade-law:
Closes #244 
Added a ternary operator to check if upcomingEvents.eventsCards is empty or not.

### DIFF
--- a/components/events/EventsPageMainContent.tsx
+++ b/components/events/EventsPageMainContent.tsx
@@ -28,7 +28,11 @@ const EventsPageMainContent: React.FC<Props> = ({ upcomingEvents }) => {
 
   return (
     <div className={mainContainer}>
-      <CoreTypography variant="h2">Upcoming Volunteer Events</CoreTypography>
+      <CoreTypography variant="h2">
+        {upcomingEvents.eventCards.length > 0
+          ? "Upcoming Volunteer Events"
+          : ""}
+      </CoreTypography>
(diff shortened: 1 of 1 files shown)


components/events/EventsPageMainContent.tsx
```diff
@@ -28,7 +28,11 @@ const EventsPageMainContent: React.FC<Props> = ({ upcomingEvents }) => {
 
   return (
     <div className={mainContainer}>
-      <CoreTypography variant="h2">Upcoming Volunteer Events</CoreTypography>
+      <CoreTypography variant="h2">
```
PC98, line 31: Can we do this for the component themselves, including the corresponding `EventsPageEventList` as opposed to just the text? Also consider using `&&`:
```javascript
{upcomingEvents.eventCards.length &&
<>
<CoreTypography / *whatever */>Upcoming Volunteer Events</CoreTypography> 
<div className={listContainer}>
<EventsPageEventList /*whatever*/ />
</div>
</>}
```

PC98 (approved):

PR #261, opened 2020-10-09
return array of coordinates given array of address string

Fattimo:


Fattimo:
image

server/actions/events.ts
```diff
@@ -149,3 +150,26 @@ export async function getAllEventIds(): Promise<string[]> {
...
+  const client = new Client({});
+
+  return Promise.all(
+    cities.map(async city => {
+      const geocode = await client.geocode({
+        params: {
+          address: city, // space delineated street address of location
+          key: config.googleMapsKey === undefined ? "" : config.googleMapsKey
+        },
+        timeout: 1000 // milliseconds
```
PC98, line 164: Question: What does this do? Do we have to specify it?

server/actions/events.ts
```diff
@@ -149,3 +150,26 @@ export async function getAllEventIds(): Promise<string[]> {
...
+
+export function getCityPolygonCoordinates(cities: string[]) {
+  const client = new Client({});
+
+  return Promise.all(
+    cities.map(async city => {
+      const geocode = await client.geocode({
+        params: {
+          address: city, // space delineated street address of location
+          key: config.googleMapsKey === undefined ? "" : config.googleMapsKey
```
PC98, line 162: Just do `key: config.googleMapsKey!` and put a `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion` on the line before.

server/actions/events.ts
```diff
@@ -149,3 +150,26 @@ export async function getAllEventIds(): Promise<string[]> {
...
+        },
+        timeout: 1000 // milliseconds
+      });
+
+      const location = geocode.data.results[0].geometry.location;
```
PC98, line 167: So since we are looking for polygon coordinates, we will have to do something a bit more involved.

We will be feeding the output of this method directly to our `$geoWithin` query that will use the `MultiPolygon` type: https://docs.mongodb.com/manual/reference/geojson/#geojson-multipolygon (note: I think `MultiPolygon` should work, but this is something worth testing), so let's make sure the return type is as expected. As an example, if the input is `["X", "Y"]`, I would expect the output to be the 4 corners of each of the cities:
```javascript
[
     [ [ [ -73.958, 40.8003 ], [ -73.9498, 40.7968 ], [ -73.9737, 40.7648 ], [ -73.9814, 40.7681 ] ] ],
     [ [ [ -73.958, 40.8003 ], [ -73.9498, 40.7968 ], [ -73.9737, 40.7648 ], [ -73.958, 40.8003 ] ] ]
]
```

PC98 (commented):
Let's delete the `package-lock.json` file. Remember, we are using `yarn` and not `npm`. After deleting this file, you can simply do `yarn install` and that will automatically fix conflicts in your `yarn.lock` file.

server/actions/events.ts
PC98: It might also be helpful to explicitly restrict ourselves to the US, `components=country:US`, see: https://developers.google.com/maps/documentation/geocoding/overview#component-filtering
The front-end also imposes a similar restriction - it will not allow users to search for a non-US city.

PC98: One last thing: the `viewport` and/or `bounds` will probably be enough to give us the polygon coordniates: https://developers.google.com/maps/documentation/geocoding/overview#GeocodingResponses

Using the `northeast` and the `southwest` points, you can figure out the coordinates for the `northwest` and `southeast` points (assume each city is a perfect rectangle) and thus you will 4 corners of each city represented. Feel free to use your own intuition for if we should use `bounds` or `viewport` though!

server/actions/events.ts
Fattimo: Nope, this is how the sample code blocks from Google did it online. It just aborts the request if it takes longer than 1000ms

PC98 (commented):
You will have to try deleting `package-lock.json` again, and then run `yarn install`, that should fix the conflicts.

server/actions/events.ts
```diff
@@ -149,3 +150,36 @@ export async function getAllEventIds(): Promise<string[]> {
...
+          key: config.googleMapsKey!
+        }
+      });
+
+      const viewport = geocode.data.results[0].geometry.viewport;
+
+      const north = viewport.northeast.lng;
+      const east = viewport.northeast.lat;
+      const south = viewport.southwest.lng;
+      const west = viewport.southwest.lat;
```
PC98, line 173: Could you verify if `viewport.northwest` and `viewport.southeast` are actually returned by the request? I know that `viewport.northeast` and `viewport.southwest` are returned for sure, see sample response: https://developers.google.com/maps/documentation/geocoding/start#geocoding-request-and-response-latitudelongitude-lookup

So you will have to compute the other two using these two (lmk if you need help with this but because we are assuming rectangular bounds, this should be straightforward).

Fattimo:
> You will have to try deleting `package-lock.json` again, and then run `yarn install`, that should fix the conflicts.

I deleted it and ran yarn install, but the file isn't changing at all. Is there another solution to this? Like maybe removing the package from the repo, then reinstalling it through yarn?

PC98 (approved):

PR #265, opened 2020-10-10
Implement 'Volunteer Events Near You' list

### LINKED ISSUE #242: [VMS] Implement the "Volunteer Events Near You" list

matthewbrown77:
Closes #242

### DIFF
--- a/components/events/EventsPageMainContent.tsx
+++ b/components/events/EventsPageMainContent.tsx
@@ -1,11 +1,15 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import { makeStyles } from "@material-ui/core";
+import CoreTypography from "@core/typography";
 
 import EventsPageEventList from "./EventsPageEventList";
-import { getUpcomingEvents } from "requests/events";
-import { DatePaginatedEventCards } from "utils/types";
+import { getUpcomingEvents, getNearestEvents } from "requests/events";
+import { usePosition } from "./usePosition";
 
-import CoreTypography from "@core/typography";
+import {
+  DatePaginatedEventCards,
+  LocationPaginatedEventCards
+} from "utils/types";
 
 const useStyles = makeStyles({
   mainContainer: {
@@ -16,6 +20,9 @@ const useStyles = makeStyles({
   },
   listContainer: {
     paddingTop: 24
+  },
+  nearestEventsContainer: {
+    marginTop: 60
   }
 });
 
@@ -24,7 +31,31 @@ interface Props {
 }
 
 const EventsPageMainContent: React.FC<Props> = ({ upcomingEvents }) => {
-  const { mainContainer, listContainer } = useStyles();
+  const { mainContainer, listContainer, nearestEventsContainer } = useStyles();
+
+  const { position, error } = usePosition();
+
+  const [nearestEvents, setNearestEvents] = useState<
+    LocationPaginatedEventCards
+  >();
+
+  useEffect(() => {
+    const fetchNearestEvents = async () => {
+      if (position) {
+        const result = await getNearestEvents({
+          lat: position.coords.latitude,
+          long: position.coords.longitude,
+          page: 0,
+          totalCount: -1,
+          isLastPage: false
+        });
+
+        setNearestEvents(result);
+      }
+    };
+
+    void fetchNearestEvents();
+  }, [position, error]);
 
   return (
     <div className={mainContainer}>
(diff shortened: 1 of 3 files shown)


matthewbrown77:
nearest_events

components/events/usePosition.ts
```diff
@@ -0,0 +1,25 @@
...
+  useEffect(() => {
+    if (!navigator || !navigator.geolocation) {
+      setError("Geolocation is not supported");
+      return;
+    }
+
+    navigator.geolocation.getCurrentPosition(onChange, onError);
+  }, []);
+
+  return { position, error };
```
matthewbrown77, line 24: @PC98 The resources you linked were very helpful for this hook. Thanks!

components/events/EventsPageMainContent.tsx
```diff
@@ -42,6 +51,29 @@ const EventsPageMainContent: React.FC<Props> = ({ upcomingEvents }) => {
           }
         />
       </div>
+      {nearestEvents && nearestEvents.eventCards.length > 0 && (
```
matthewbrown77, line 54: This PR won't pass all the automated checks because the `upcomingEvents` container does do a null check yet. I thought about adding it here but I saw there was another ticket for it.

requests/events.ts
```diff
@@ -22,7 +22,7 @@ export const getNearestEvents = async (
   locationPageInformation: LocationPageInformation
 ) =>
   fetchRequestWithPayloadResponse<LocationPaginatedEventCards>(
-    config.apis.getUpcomingEvents,
+    config.apis.getNearestEvents,
```
PC98, line 25: Good catch! 😅

components/events/EventsPageUnfiltered.tsx
```diff
@@ -12,9 +18,36 @@ const EventsPageUnfiltered: React.FC<Props> = ({
...
+          totalCount: -1,
+          isLastPage: false
+        });
+
+        setNearestEventsData(result);
+      }
+    };
+
+    void fetchNearestEvents();
+  }, [position, error]);
```
PC98, line 43: What are our thoughts on moving this code to `EventsPageMainContent` directly?

components/events/EventsPageMainContent.tsx
```diff
@@ -16,15 +18,22 @@ const useStyles = makeStyles({
...
     paddingTop: 24
+  },
+  nearestEventsContainer: {
+    marginTop: 60
   }
 });
 
 interface Props {
-  upcomingEvents: DatePaginatedEventCards;
+  upcomingEvents: DatePaginatedEventCards | null;
```
PC98, line 28: I don't think you need to change the type of `upcomingEvents` here

components/events/EventsPageMainContent.tsx
```diff
@@ -16,15 +18,22 @@ const useStyles = makeStyles({
...
+  nearestEvents: LocationPaginatedEventCards | undefined;
```
PC98, line 29: If we follow the suggestion, we can avoid having to pass `nearestEvents` as props.

components/events/EventsPageMainContent.tsx
PC98: Replied above

PC98 (commented):
Let me know if you need help resolving conflicts with the `develop` branch, but this looks great!

components/events/EventsPageUnfiltered.tsx
matthewbrown77: It does make probably make more sense. No reason to pass props when we don't have to. I just though the file was getting a little bloated already and didn't want to add more complexity.

components/events/EventsPageMainContent.tsx
matthewbrown77: We still have to do this extra check since we aren't setting a default object for `nearestEvents`. We could add one with dummy data to remove this check, but it seemed cleaner to just let this be.

components/events/EventsPageUnfiltered.tsx
PC98: Tbh we might eventually have to move this to a separate component altogether, perhaps some child component that `EventsPageMainContent` renders. We will want to make sure that when a user interacts with the location list, it doesn't re-render the upcoming list and vice-versa.

components/events/EventsPageMainContent.tsx
PC98: Yep the check for `nearestEvents` makes sense, I was referring to your comment about `upcomingEvents`.

As a future nit, it's always more helpful to be explicit:
```javascript
const [nearestEvents, setNearestEvents] = useState<
     LocationPaginatedEventCards | null
   >(null);
```
Right now it's not obvious from reading the code that `nearestEvents`'s type is actually `LocationPaginatedEventCards | undefiend`

PC98 (approved):

PR #279, opened 2020-10-13
get loc from query, filter by city as well

Fattimo:
read query params to get loc params, change filter method to filter by city as well

changed return value of getCityPolygonCoordinates to be a closed loop


server/actions/events.ts
```diff
@@ -181,7 +196,8 @@ export function getCityPolygonCoordinates(cities: string[]) {
           [north, east],
           [south, east],
           [south, west],
-          [north, west]
+          [north, west],
+          [north, east]
```
PC98: Why do we need an extra `[north, east]`?

server/actions/events.ts
```diff
@@ -120,18 +120,33 @@ export async function getNearestEventsCardDataCount({
...
   const idsWithCause = await Nonprofit.find(
     {
       cause: { $in: causes }
     },
     "nonprofitId"
   );
 
+  console.log(idsWithCause);
+
+  console.log(cities);
```
PC98: remove `console.log`s

server/actions/events.ts
```diff
@@ -120,18 +120,33 @@ export async function getNearestEventsCardDataCount({
...
+    nonprofitId: { $in: idsWithCause.map(r => r["_id"]) },
+    "address.location": {
+      $geoWithin: {
+        $geometry: {
+          type: "MultiPolygon",
+          coordinates: bounds
+        }
+      }
+    }
   }).limit(5);
```
PC98: Question: What would happen if exactly one of the two or both, `cities` and `causes`, is an empty array? Does the query behave as expected? If it's hard to test this function with multiple args, feel free to convert them to an object.

pages/events/index.tsx
```diff
@@ -61,7 +61,11 @@ export const getServerSideProps = async (
     };
   } else {
     const query = getFilterValuesInQuery(context.query, "cause");
-    /*upcomingEventsFirstPageData = */ await getByCausesEventsCardData(query);
+    const cities = getFilterValuesInQuery(context.query, "location");
+    /*upcomingEventsFirstPageData = */ await getByCausesEventsCardData(
+      query,
```
PC98: nit: consider renaming `query` to `causes`.

server/actions/events.ts
Fattimo: This is required for a geospatial query, as it makes a closed loop. https://stackoverflow.com/questions/47665844/mongodb-geospatial-query-geowithin-got-error-loop-is-not-closed

PC98: Might be worth adding a comment to explain this!

server/actions/events.ts
Fattimo: With empty cities, there is an error within this method as a geoquery cannot have zero elements.

With empty causes, the empty array is returned.

I think the way to go is to make them default to arrays filled with every single possible option (I like this one more; unsure of how frontend is currently coded, but I expect that there are variables somewhere that already have this information), or use switch cases to change out the filter in `find()` based on what has been passed into the method (better for performance I think).

PC98: > use switch cases to change out the filter in find() based on what has been passed into the method (better for performance I think).

Yes, let's use this approach. You code can just have one `Event.find`, but the object being passed to it can vary. You can make use of the spread operator here:
```javascript
let findQuery = {}; // initially empty -- if this method is passed empty arrays, we don't have any filtering criteria
if (causes.length) {
  // non-empty causes, do some stuff 
  findQuery = {...findQuery, nonprofitId: { $in: idsWithCause.map(r => r["_id"]) } };
}
if (cities.length) {
 // non-empty cities, do some stuff
 findQuery = {...findQuery, "address.location": /*whatever*/ }
}
// continue
```

PC98 (approved):

PR #280, opened 2020-10-13
Create generic list for cards

### LINKED ISSUE #215: [VMS] Decouple pagination list logic and UI from event-specific implementation
* The `EventsList` component works great! But it's highly coupled to displaying data of type `EventDisplayInfo` and can only render `EventCardLarge` and `EventCardGlimmerLarge`.
* Let's create a new list component which does not have the above constraints but retains all the other features of `EventsList` like the next and the back button icons.
* Essentially, this will be a generic component. As an example of creating a generic component, see "Implementing a fully working table" here: https://wanago.io/2020/03/09/functional-react-components-with-generic-props-in-typescript/
* We might have to use child render props: https://fettblog.eu/typescript-react/render-props/#children-render-props, let's talk about this on Slack after you have given this issue some thought.
* This task is important so that the "Volunteer for a Cause" list can also use pagination.

migueldlr:
Closes #215 

We create `EventsPageCardList`, which displays paginated data. In creating `EventsPageCardList`, we also refactor `EventsPageEventList` to use `EventsPageCardList`.

Other changes:
- rename `eventCards` to `cards`


components/events/EventsPageCardList.tsx
```diff
@@ -0,0 +1,183 @@
...
+    newPage: number
+  ) => Promise<PageInformation & { cards: CardData[] }>;
+  renderCard: (c: CardData | null) => JSX.Element;
+}
+
+const DEFAULT_ROW_SIZE = 4;
+const MARGIN_ADJUSTMENT = 24;
+const CARD_WIDTH = 283;
+
+const EventsPageCardList = <CardData,>({
```
PC98: Can we make the naming even more generic (no mention of "event"), and put it in `components/` (possibly even `components/core/list` with a name that indicates the component supports pagination)?


I found out today that the core team is interested in using your implementation for their "New Non-profits on our Platform" list! 👀💯

components/events/EventsPageCardList.tsx
```diff
@@ -0,0 +1,183 @@
...
+  paginatedCardsData: PageInformation & { cards: CardData[] };
+  fetchCards: (
+    newPage: number
+  ) => Promise<PageInformation & { cards: CardData[] }>;
+  renderCard: (c: CardData | null) => JSX.Element;
+}
+
+const DEFAULT_ROW_SIZE = 4;
+const MARGIN_ADJUSTMENT = 24;
+const CARD_WIDTH = 283;
```
PC98: Will some of these values have to be taken as props, now that users specify their own card render component?

It would be super nice if we could be `CARD_WIDTH` agnostic, but I imagine that will require some serious CSS work so I am okay with taking width as a prop.

components/events/EventsPageCardList.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+const EventsPageCardList = <CardData,>({
+  paginatedCardsData,
+  fetchCards,
+  renderCard
+}: Props<CardData>) => {
+  const classes = useStyles();
+
+  const [events, setEvents] = useState(paginatedCardsData.cards);
```
PC98: nit: remove all mentions of "events",  so that someone else reading the code ins't confused

components/events/EventsPageCardList.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  // Number of elements displayed
+  const [rowSize, setRowSize] = useState(DEFAULT_ROW_SIZE);
+  const [maxElem, setMaxElem] = useState(-1);
+  const [loading, setLoading] = useState(false);
+
+  const containerRef = useRef<HTMLDivElement>(null);
+  const pageRef = useRef(paginatedCardsData.page);
+
+  // Lock calls to getEvents if one is currently in progress
```
PC98: nit: update comment to not use getEvents

components/events/EventsPageCardList.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const resizeTimeoutRef = useRef<number>();
+
+  useEffect(() => {
+    if (first + rowSize > events.length && !fetchingRef.current) {
+      void (async () => {
+        fetchingRef.current = true;
+        setLoading(true);
+
+        const {
+          cards: newEvents,
```
PC98: `newCards`

components/events/EventsPageCardList.tsx
```diff
@@ -0,0 +1,183 @@
...
+          isLastPage
+        } = await fetchCards(pageRef.current + 1);
+
+        pageRef.current = newPage;
+
+        if (isLastPage) {
+          setMaxElem(events.length + newEvents.length);
+        }
+
+        setEvents(prevEvents => [...prevEvents, ...newEvents]);
```
PC98: here as well

components/events/EventsPageCardList.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const nextPage = () => {
+    setFirst(first + rowSize);
+  };
+
+  const prevPage = () => {
+    setFirst(Math.max(first - rowSize, 0));
+  };
+
+  // holds all the elements currently displayed - gets rendered as
+  // EventsPageEventCard, while null gets rendered as EventsPageEventCardGlimmer
```
PC98: update comment to not have `EventsPageEventCard` and `EventsPageEventCardGlimmer`

components/events/EventsPageCardList.tsx
```diff
@@ -0,0 +1,183 @@
...
+    nextButtonContainer: {
+      marginLeft: -56,
+      borderRadius: "50%"
+    },
+    prevButtonContainer: {
+      marginLeft: -24,
+      marginRight: -24,
+      borderRadius: "50%",
+      position: "relative",
+      zIndex: 1
```
PC98: Will some of these styles still be applicable if a custom card component is rendered, particularly those pertaining to the positioning of the buttons? We should consider adding props for styles as well.

components/events/EventsPageCardList.tsx
```diff
@@ -0,0 +1,183 @@
...
+      marginRight: 32
+    }
+  })
+);
+
+interface Props<CardData> {
+  paginatedCardsData: PageInformation & { cards: CardData[] };
+  fetchCards: (
+    newPage: number
+  ) => Promise<PageInformation & { cards: CardData[] }>;
```
PC98: To avoid doing this `PageInformation & { cards: CardData[] }` multiple times and to create a re-usable type, you can create a generic type in `uitls/types.ts`:
```javascript
type PaginatedCards<T> = PageInformation & { cards: T[] }
```

components/events/EventsPageEventList.tsx
```diff
@@ -1,201 +1,44 @@
...
-          >
-            <ChevronRightIcon />
-          </IconButton>
-        </div>
-      )}
-    </div>
+    <EventsPageCardList
+      paginatedCardsData={paginatedEventCardsData}
+      fetchCards={getMoreEvents}
+      renderCard={cardData =>
```
PC98: Question on API design:
As someone using `EventsPageCardList`/this generic component without being familiar with the internal implementation details, it might be weird that `renderCard` receives `null`able `cardData`, even though all the card data I am passing is non-null.

It might be better to have two props - `renderCard`, which is a function that always receives non-null data, and `loadingCardComponent` (or whatever appropriate name you can think of) which is simply a `React.ReactNode` or `JSX.Element`.

Optionally, you can then also clean-up some code in `EventsPageCardList`:
```javascript
 const display = events.slice(first, first + rowSize).map(renderCard);
   const hasNext = maxElem == -1 || first + rowSize < maxElem;
   if (hasNext) {
     // pad the display with loading card components
     while (display.length < rowSize) {
       display.push(loadingCardComponent);
     }
   }
```

PC98 (commented):
We will be using this component a lot 🔥thanks for working on making it more flexible

PC98 (approved):

PR #298, opened 2020-10-16
Formatted date and times for event cards

### LINKED ISSUE #245: [VMS] Text representation of an event's `startTime` and `endTime`
* Using the `startDate`, `endDate` (and potentially `duration`) fields of the passed in `EventCardData` object, `EventsPageEventCard.tsx` should compute and display a string representing the start and end time of the event, as shown in the Figma designs.
* Try doing this task without adding a date library as a dependency, this might be helpful: https://github.com/you-dont-need/You-Dont-Need-Momentjs
* However if you must add a date library to help you, consider using date-fns or dayjs.

jade-law:
Closes #245 
Added a function that would format the date and times for the event cards.

### DIFF
--- a/components/events/EventsPageEventCard.tsx
+++ b/components/events/EventsPageEventCard.tsx
@@ -104,7 +104,7 @@ const EventsPageEventCard: React.FC<Props> = ({ eventCardData, onClick }) => {
       endFormatted = `${endMonth} ${endDateString}, ${endTime}`;
     }
     return `${startFormatted} - ${endFormatted}`;
-  }, [eventCardData.startDate, eventCardData.endDate]);
+  }, [eventCardData]);
 
   return (
     <FocusVisibleOnly onClick={onClick}>
--- a/components/events/EventsPageEventCard.tsx
+++ b/components/events/EventsPageEventCard.tsx
@@ -78,6 +78,24 @@ const EventsPageEventCard: React.FC<Props> = ({ eventCardData, onClick }) => {
     truncate
   } = useStyles();
 
+  const formatDate = (event: EventCardData) => {
+    const startDate = new Date(event.startDate);
+    const startDateFormatted = startDate.toLocaleString("en-US", {
+      month: "short",
+      day: "numeric",
+      hour: "numeric",
+      minute: "numeric",
+      hour12: true
+    });
+    const endDate = new Date(event.endDate);
+    const endDateFormatted = endDate.toLocaleString("en-US", {
+      hour: "numeric",
+      minute: "numeric",
+      hour12: true
+    });
+    return startDateFormatted + "-" + endDateFormatted;
+  };
+
   return (
     <FocusVisibleOnly onClick={onClick}>
       <div className={cardContainer}>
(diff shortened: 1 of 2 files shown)


components/events/EventsPageEventCard.tsx
```diff
@@ -78,6 +78,24 @@ const EventsPageEventCard: React.FC<Props> = ({ eventCardData, onClick }) => {
...
+      hour: "numeric",
+      minute: "numeric",
+      hour12: true
+    });
+    const endDate = new Date(event.endDate);
+    const endDateFormatted = endDate.toLocaleString("en-US", {
+      hour: "numeric",
+      minute: "numeric",
+      hour12: true
+    });
```
PC98: You might have to consider an edge case here, if the `endDate` day is not the same as `startDate` day. For instance, you might have an event that begins on Jun 27th, but ends on Jun 28th. In that case, let's format `endDate` similar to `startDate` so that "Jun 28" also gets displayed.

components/events/EventsPageEventCard.tsx
```diff
@@ -78,6 +78,24 @@ const EventsPageEventCard: React.FC<Props> = ({ eventCardData, onClick }) => {
...
+    return startDateFormatted + "-" + endDateFormatted;
```
PC98: Nit: consider using a [template string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals), and you might also need some space around the "-":
```javascript
return `${startDateFormatted} - ${endDateFormatted}`;
```

components/events/EventsPageEventCard.tsx
```diff
@@ -78,6 +78,24 @@ const EventsPageEventCard: React.FC<Props> = ({ eventCardData, onClick }) => {
...
 
+  const formatDate = (event: EventCardData) => {
+    const startDate = new Date(event.startDate);
+    const startDateFormatted = startDate.toLocaleString("en-US", {
+      month: "short",
+      day: "numeric",
+      hour: "numeric",
+      minute: "numeric",
+      hour12: true
+    });
```
PC98: Thanks for figuring out a way to do this without adding a new dependency! 🔥

components/events/EventsPageEventCard.tsx
```diff
@@ -78,6 +78,24 @@ const EventsPageEventCard: React.FC<Props> = ({ eventCardData, onClick }) => {
     truncate
   } = useStyles();
 
+  const formatDate = (event: EventCardData) => {
```
PC98: Consider using `useMemo` here so that you compute a value directly as opposed to declaring a function:
```javascript
const formattedDate = useMemo(() => {
  const startDate = new Date(eventCardData.startDate);
  // your code as usual
  return /*SOMETHING*/;
  }, [eventCardData]);
```

PC98:
I noticed that the Figma designs also include the ordinal, so "Jun 27**th**" instead of just "Jun 27".

To achieve this, try taking this approach: https://stackoverflow.com/questions/15397372/javascript-new-date-ordinal-st-nd-rd-th
You can make a new method, put it in `utils/util.ts` perhaps, and call it something like `getNumberOrdinal` (that takes in a number and returns the ordinal string, so `getNumberOrdinal(27)` would return `"th"`)  or `getNumberWithOrdinal` (that can take in a string number and return the string with the ordinal, so `getNumberWithOrdinal("27")` would return `"27th"`).

PC98 (approved):

PC98:
Awesome!

PR #299, opened 2020-10-18
Create and add EventsPageCauseList

### LINKED ISSUE #247: [VMS] UI for "Volunteer for a Cause" list
* Add this list to `EventsPageMainContent.tsx`.
* Use the list component added as part of #215.
* You can access the list of causes like how `EventsPageSidebarComponent` does it.
* The card and the glimmer components have already been implemented for you; see `EventsPageCauseCard` and `EventsPageCauseCardGlimmer`.
* You should think about that the `onClick` prop that you pass to `EventsPageCauseCard` should be for each cause.

migueldlr:
Closes #247 

Creates `EventsPageCauseList.tsx` which renders a `CardPaginationList`. Also modifies `CardPaginationList` to handle a static list of items.


components/events/EventsPageMainContent.tsx
```diff
@@ -103,6 +105,12 @@ const EventsPageMainContent: React.FC<Props> = ({ upcomingEvents }) => {
           </div>
         </>
       )}
+      <div className={nearestEventsContainer}>
+        <CoreTypography variant="h2">Volunteer For a Cause</CoreTypography>
+        <div className={listContainer}>
+          <EventsPageCauseList paginatedCauseCardsData={CAUSE_CARDS} />
```
PC98: nit: you could avoid taking props and avoid having a `utils/causes.ts`; in `EventsPageCauseList.tsx` you can directly create `CAUSE_CARDS`, outside the functional component.

utils/types.ts
```diff
@@ -64,6 +64,14 @@ export type PaginatedEventCards = PageInformation & {
 
 export type PaginatedCards<CardData> = PageInformation & { cards: CardData[] };
 
+export type CauseCardData = {
+  cause: string;
+  imagePath: string;
+  filterValue: string;
+};
+
+export type PaginatedCauseCards = PageInformation & { cards: CauseCardData[] };
```
PC98: So since we created a generic type, `PaginatedCards`, instead of `PageInformation & { cards: CauseCardData[] };` you can just do `PaginatedCards<CauseCardData>`.

Optionally, also consider refactoring the `PaginatedEventCards` type, and instead of `PaginatedCards<CardData>`, use `PaginatedCards<T>` to make it more obvious that this is a generic type/encourage re-usability.

components/events/EventsPageCauseList.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const EventsPageCauseList: React.FC<Props> = ({ paginatedCauseCardsData }) => {
+  const { put } = useRouterQueryParamsState("cause");
+
+  return (
+    <CardPaginationList
+      paginatedCardsData={paginatedCauseCardsData}
+      cardGlimmer={<EventsPageCauseCardGlimmer />}
+      renderCard={cardData => (
+        <EventsPageCauseCard
```
PC98, line 32: nit: consider refactoring `EventsPageCauseCard` to work directly with `CauseCardData`, just like we refactored `EventsPageEventCard` to use `EventCardData`

PC98 (approved):
💯

PR #301, opened 2020-10-18
added filtering by startDate

### LINKED ISSUE #246: [VMS] Return events filtered by startDate
* Similar to #217, which added functionality to query for events by causes based on the URL query parameters, we will now implement similar functionality to filter for events whose startDate <= any one of select "time" filters.
* Modify the existing `getByCausesEventsCardData()` to handle this functionality. You should also make sure to:
- [x] Rename the function to something more appropriate, `getFilteredEvents` or something.
- [x] Modify `causes` to be of type `FilterValue<"cause">[]`
- [x] The `times` argument you add should be of type `FilterValue<"time">[]`

stevenkchen:
added filtering by start date. Go through all the times passed in and then calculate the time ranges. Use a or statement for the mongo find.

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -222,9 +222,7 @@ export async function getByFilteredEventsCardData(
       $or: timeFilters
     };
   }
-  console.log(JSON.stringify(findQuery));
   const result = await Event.find(findQuery).limit(5);
-  console.log(result);
   return result.map(r => r.toJSON()) as EventCardDataType[];
 }
 
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -179,14 +179,14 @@ export async function getByFilteredEventsCardData(
           break;
         }
         case "WEEKEND": {
-          const offset = startTime.getDay() == 0 ? 0 : 6 - startTime.getDay();
+          const offset = startTime.getDay() == 0 ? -1 : 6 - startTime.getDay();
           startTime.setDate(startTime.getDate() + offset);
           endTime = new Date(startTime);
           endTime.setDate(startTime.getDate() + 2);
           break;
         }
         case "NWEEKEND": {
-          const offset = startTime.getDay() == 0 ? 0 : 6 - startTime.getDay();
+          const offset = startTime.getDay() == 0 ? -1 : 6 - startTime.getDay();
           startTime.setDate(startTime.getDate() + offset + 7);
           endTime = new Date(startTime);
           endTime.setDate(startTime.getDate() + 2);
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -159,14 +159,13 @@ export async function getByFilteredEventsCardData(
   }
 
   if (times.length) {
-    let timeFilters: any[] = [];
-    for (const time of times) {
+    const timeFilters = times.map(time => {
       let startTime = new Date();
       startTime.setHours(0);
       startTime.setMinutes(0);
       startTime.setSeconds(0);
       startTime.setMilliseconds(0);
-      let endTime = null;
+      let endTime;
       switch (time) {
         case "TODAY": {
           endTime = new Date(startTime);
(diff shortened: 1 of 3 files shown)


server/actions/events.ts
```diff
@@ -156,8 +158,69 @@ export async function getByCausesEventsCardData(
     };
   }
 
+  if (times.length) {
+    let timeFilters: any[] = [];
```
PC98: To avoid `any`, declare this as `const timeFilters = [];` and perform the suggested change on lines 210 to 216

server/actions/events.ts
```diff
@@ -156,8 +158,69 @@ export async function getByCausesEventsCardData(
...
+          break;
+        }
+      }
+      const tempQuery = {
+        startDate: {
+          $gte: startTime,
+          $lt: endTime
+        }
+      };
+      timeFilters = [...timeFilters, tempQuery];
```
PC98:
```javascript
timeFilters.push({
  startDate: {
    $gte: startTime,
    $lt: endTime
  }
});
```

server/actions/events.ts
```diff
@@ -156,8 +158,69 @@ export async function getByCausesEventsCardData(
...
+          break;
+        }
+        case "NWEEK": {
+          startTime = new Date();
+          startTime.setDate(startTime.getDate() + 7);
+          endTime = new Date(startTime);
+          endTime.setDate(startTime.getDate() + 7);
+          break;
+        }
+        case "WEEK": {
```
PC98, line 202: After this `case`, add an exhaustive check in a `default`. This will let TS tell us if we have taken into account all variants of `time` or not:
```javascript
default: {
  const _exhaustiveCheck: never = time;
  return _exhaustiveCheck;
}
```

server/actions/events.ts
```diff
@@ -156,8 +158,69 @@ export async function getByCausesEventsCardData(
...
+          break;
+        }
+        case "TOMORROW": {
+          startTime.setDate(startTime.getDate() + 1);
+          endTime = new Date(startTime);
+          endTime.setDate(startTime.getDate() + 1);
+          break;
+        }
+        case "WEEKEND": {
+          const offset = 6 - startTime.getDay();
```
PC98: We should think about what happens when `startTime` is Sunday. I think `offset` would be 6 - 0 = 6. In that case, `startTime.setDate(startTime.getDate() + offset);` would change `startTime` to next Saturday, but we want it to be this weekend/saturday. So maybe add an if-else to check if `offset` is 6.

PC98: Same comment applies to "NWEEKEND"

server/actions/events.ts
```diff
@@ -156,8 +158,69 @@ export async function getByCausesEventsCardData(
     };
   }
 
+  if (times.length) {
+    let timeFilters: any[] = [];
+    for (const time of times) {
```
PC98: As an FYI, you can also use JavaScript's `.map()` function 🙂
```javascript
const timeFilters = times.map(time => {
  // your code
  return {
    startDate: {
      $gte: startTime,
      $lt: endTime
    }
  };
});
```

server/actions/events.ts
```diff
@@ -156,8 +158,69 @@ export async function getByCausesEventsCardData(
...
+      let startTime = new Date();
+      startTime.setHours(0);
+      startTime.setMinutes(0);
+      startTime.setSeconds(0);
+      startTime.setMilliseconds(0);
+      let endTime = null;
```
PC98: nit: you can drop the assignment to `null`: `let endTime;`

server/actions/events.ts
```diff
@@ -156,8 +158,71 @@ export async function getByCausesEventsCardData(
...
+          endTime = new Date(startTime);
+          endTime.setDate(startTime.getDate() + 1);
+          break;
+        }
+        case "WEEKEND": {
+          const offset = startTime.getDay() == 0 ? 0 : 6 - startTime.getDay();
+          startTime.setDate(startTime.getDate() + offset);
+          endTime = new Date(startTime);
+          endTime.setDate(startTime.getDate() + 2);
+          break;
```
PC98: @jade-law Thanks for iterating on this!
Though I think this still won't work for Sunday, that is, when `startTime.getDay() == 0`. It seems `startTime` will get set to Sunday 12:00 AM and `endTime` will get set to Tuesday:
Screen Shot 2020-10-18 at 8 05 30 PM

cc: @schen2425

PC98: Though I think `const offset = startTime.getDay() == 0 ? -1 : // rest of your code` should work

PC98 (approved):

PR #311, opened 2020-10-19
Create events listing header

### LINKED ISSUE #294: [VMS] UI for EventsListing page header
* Look at the Figma design for the EventsListing page. You will be implementing just the UI portion (no back-end data integration) for the header which contains the text "380 Results" and the "Sort by" dropdown.
* You can create a new component that gets rendered by `EventsPageFiltered.tsx`. As an example, see how `EventsPageUnfiltered.tsx` renders `EventsPageMainContent`.
* You can use material-ui's `Select` component. For reference, `DonationPageHeaderSelect.tsx` uses it.
* You should use the `usePosition` hook. If the user selects "Closest to you", only then we should ask the user for their location. Though remember you can't use a hook conditionally. So you might have to either modify `usePosition` to control when the `useEffect` inside it gets called or think of some other clever way.
* If the user has denied sharing their location information with us, we should prevent them from selecting "Closest to you"/switch to the other option. In this case, we can show a toast to the user (https://material-ui.com/components/snackbars/#snackbar) or a simple JS window alert  to tell them they can't select the "Closest to you" options since we can't their location.
* For now, we **won't** update the URL through our dropdown.

matthebrown:
Closes #294

### DIFF
--- a/components/events/header/EventsPageFilteredHeaderAlert.tsx
+++ b/components/events/header/EventsPageFilteredHeaderAlert.tsx
@@ -1,0 +1,39 @@
+import React, { useState } from "react";
+
+import { IconButton, Snackbar } from "@material-ui/core";
+import { Alert } from "@material-ui/lab";
+
+import { XIcon } from "@core/icons";
+
+const EventsPageFilteredHeaderAlert: React.FC = () => {
+  const [open, setOpen] = useState(true);
+
+  return (
+    <div>
+      <Snackbar
+        anchorOrigin={{ vertical: "top", horizontal: "right" }}
+        open={open}
+      >
+        <Alert
+          severity="warning"
+          action={
+            <IconButton
+              aria-label="close"
+              color="inherit"
+              size="small"
+              onClick={() => {
+                setOpen(false);
+              }}
+            >
+              <XIcon fontSize="inherit" />
+            </IconButton>
+          }
+        >
+          You must share your location to sort by &quot;Closest to you&quot;
+        </Alert>
+      </Snackbar>
+    </div>
+  );
+};
+
+export default EventsPageFilteredHeaderAlert;
--- a/components/events/header/EventsPageFilteredHeaderAlert.tsx
+++ b/components/events/header/EventsPageFilteredHeaderAlert.tsx
@@ -1,38 +1,22 @@
 import React, { useState } from "react";
 
-import { IconButton, Snackbar } from "@material-ui/core";
+import { Snackbar } from "@material-ui/core";
 import { Alert } from "@material-ui/lab";
-
-import { XIcon } from "@core/icons";
 
 const EventsPageFilteredHeaderAlert: React.FC = () => {
   const [open, setOpen] = useState(true);
 
   return (
-    <div>
-      <Snackbar
-        anchorOrigin={{ vertical: "top", horizontal: "right" }}
-        open={open}
-      >
-        <Alert
-          severity="warning"
-          action={
-            <IconButton
-              aria-label="close"
-              color="inherit"
-              size="small"
-              onClick={() => {
-                setOpen(false);
-              }}
-            >
-              <XIcon fontSize="inherit" />
-            </IconButton>
-          }
-        >
-          You must share your location to sort by &quot;Closest to you&quot;
-        </Alert>
-      </Snackbar>
-    </div>
+    <Snackbar
+      anchorOrigin={{ vertical: "bottom", horizontal: "left" }}
+      open={open}
+      autoHideDuration={8000}
+      onClose={() => setOpen(false)}
+    >
+      <Alert severity="warning">
+        You must share your location to sort by &quot;Closest to you&quot;
+      </Alert>
+    </Snackbar>
   );
 };
 
--- a/components/events/header/EventsPageFilteredHeaderSelect.tsx
+++ b/components/events/header/EventsPageFilteredHeaderSelect.tsx
@@ -1,0 +1,147 @@
+import React, { useState, useCallback, useRef } from "react";
+
+import { makeStyles, Select, MenuItem, FormControl } from "@material-ui/core";
+
+import { ChevronDownIcon } from "@core/icons";
+import CoreTypography from "@core/typography";
+import { DropdownProps } from "utils/types";
+
+import { usePosition } from "../usePosition";
+import EventsPageFilteredHeaderAlert from "./EventsPageFilteredHeaderAlert";
+
+const useStyles = makeStyles({
+  container: {
+    display: "flex",
+    color: "#FD8033",
+    paddingTop: 6
+  },
+  select: {
+    color: "#FD8033"
+  },
+  input: {
+    "&:focus": {
+      backgroundColor: "transparent"
+    },
+    paddingRight: "0 !important"
+  },
+  icon: {
+    color: "#FD8033",
+    fontSize: "0.90rem",
+    marginTop: 2.5,
+    marginLeft: 10
+  },
+  menu: {
+    display: "flex",
+    flexDirection: "column",
+    alignItems: "center",
+    justifyContent: "center",
+    border: "1px solid #F0F0F0",
+    borderRadius: 10,
+    boxSizing: "border-box",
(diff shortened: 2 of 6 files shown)


matthebrown:
new_sort
options

components/events/header/EventsPageFilteredHeaderAlert.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Alert } from "@material-ui/lab";
+
+import { XIcon } from "@core/icons";
+
+const EventsPageFilteredHeaderAlert: React.FC = () => {
+  const [open, setOpen] = useState(true);
+
+  return (
+    <div>
+      <Snackbar
```
PC98: consider using the `autoHideDuration` prop to auto-dismiss after some time (this prop accepts number of milliseconds).

components/events/header/EventsPageFilteredHeaderAlert.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { IconButton, Snackbar } from "@material-ui/core";
+import { Alert } from "@material-ui/lab";
+
+import { XIcon } from "@core/icons";
+
+const EventsPageFilteredHeaderAlert: React.FC = () => {
+  const [open, setOpen] = useState(true);
+
+  return (
+    <div>
```
PC98: do you need this outer `div`?

components/events/header/EventsPageFilteredHeaderAlert.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <Snackbar
+        anchorOrigin={{ vertical: "top", horizontal: "right" }}
```
PC98: The snackbar blocks the "Sort by" text when it appears, at least on my screen. Can we render this bottom left instead?

components/events/header/EventsPageFilteredHeaderSelect.tsx
```diff
@@ -0,0 +1,147 @@
...
+import CoreTypography from "@core/typography";
+import { DropdownProps } from "utils/types";
+
+import { usePosition } from "../usePosition";
+import EventsPageFilteredHeaderAlert from "./EventsPageFilteredHeaderAlert";
+
+const useStyles = makeStyles({
+  container: {
+    display: "flex",
+    color: "#FD8033",
```
PC98: Instead of using "#FD8033" like this, use `palette.primary.main` everywhere, see: https://github.com/GTBitsOfGood/gen-soln/blob/develop/components/events/EventsPageEventCard.tsx#L52

components/events/header/EventsPageFilteredHeaderSelect.tsx
```diff
@@ -0,0 +1,147 @@
...
+    icon,
+    menu,
+    menuRoot,
+    menuItemRoot,
+    menuItemSelected
+  } = useStyles();
+
+  const isDisabled = useRef(true);
+
+  const { position, error } = usePosition(isDisabled.current);
```
PC98: nit: lint warning for unsued `position`

components/events/header/EventsPageFilteredHeaderSelect.tsx
```diff
@@ -0,0 +1,147 @@
...
+        >
+          {items.map(({ text, value }) => (
+            <MenuItem
+              key={value}
+              value={value}
+              classes={{
+                root: menuItemRoot,
+                selected: menuItemSelected
+              }}
+              disabled={(error && value === "location") as boolean}
```
PC98: avoid the `as`: `disabled={error != null && value === "location"}`

components/events/header/EventsPageFilteredHeaderSelect.tsx
```diff
@@ -0,0 +1,147 @@
...
+      {error && <EventsPageFilteredHeaderAlert />}
+      <FormControl className={input}>
+        <CoreTypography variant="h4">Sort by</CoreTypography>
+        <Select
+          classes={{ select }}
+          value={value}
+          onChange={onChange}
+          autoWidth={true}
+          disableUnderline={true}
+          IconComponent={() => <ChevronDownIcon className={icon} />}
```
PC98, line 125: `IconComponent={ChevronDownIcon}`, since we want this to be an `elementType`. To style the icon, use `Select` `classes`'s `icon` property.

This should also make the icon clickable and flip the icon when the input menu is open.

components/events/header/EventsPageFilteredHeaderSelect.tsx
```diff
@@ -0,0 +1,147 @@
...
+    borderRadius: 10,
+    boxSizing: "border-box",
+    boxShadow: "0 5px 10px"
+  },
+  menuRoot: {
+    borderRadius: 10
+  },
+  menuItemRoot: {
+    "&$menuItemSelected, &$menuItemSelected:focus, &$menuItemSelected:hover": {
+      backgroundColor: "#FD8033",
```
PC98: I will follow-up with designers for what we want here, but for now, this looks good!

components/events/header/EventsPageFilteredHeaderSelect.tsx
```diff
@@ -0,0 +1,147 @@
...
+    fontSize: "0.90rem",
+    marginTop: 2.5,
+    marginLeft: 10
+  },
+  menu: {
+    display: "flex",
+    flexDirection: "column",
+    alignItems: "center",
+    justifyContent: "center",
+    border: "1px solid #F0F0F0",
```
PC98: use this color this way: https://github.com/GTBitsOfGood/gen-soln/blob/develop/components/events/EventsPageCauseCardGlimmer.tsx#L14

components/events/header/EventsPageFilteredHeaderSelect.tsx
```diff
@@ -0,0 +1,147 @@
...
+      setValue(event.target.value as typeof selectedValue);
+
+      if (event.target.value === "location") {
+        isDisabled.current = false;
+      }
+    },
+    []
+  );
+
+  return (
```
PC98, line 114: Let's also add a hook to undo the selection of the location dropdown if we can't get the user's location:
```javascript
useEffect(() => {
  if (error != null && value === "location") {
    setValue("participants");
  }
}, [error, value]);
```
Feel free to assign `error != null && value === "location"` to a well named variable, since you are gonna be using this value on line 136 as well.

components/events/header/EventsPageFilteredHeaderSelect.tsx
```diff
@@ -0,0 +1,147 @@
...
+const useStyles = makeStyles({
+  container: {
+    display: "flex",
+    color: "#FD8033",
+    paddingTop: 6
+  },
+  select: {
+    color: "#FD8033"
+  },
+  input: {
```
PC98: Can we add a `minWidth` value here that is large enough so that the position of the "Sort by" text doesn't change when "Closest to you" gets selected?

components/events/header/EventsPageFilteredHeaderSelect.tsx
PC98: Also, this is optional, but we can use TypeScript to ensure that `value` can actually be `location`:
1. First, remove any props from this component, and then directly above it do:
```javascript
const SORT_OPTIONS = [
  { text: "Closest to you", value: "location" },
  { text: "Most signed up", value: "participants" }
] as const;
type OptionValue = typeof SORT_OPTIONS[number]["value"];
```
2. Update `value`:
```javascript
const [value, setValue] = useState<OptionValue>("participants")
``` 
3. Update `onChange`:
```javascript
const selectedValue = event.target.value as OptionValue;
setValue(selectedValue);

if (selectedValue === "location") { // continue
```

Another added benefit of this approach is that our options live directly in this component, so someone reading this code won't have to figure out what "location" is and where it comes from.

components/events/header/EventsPageFilteredHeaderSelect.tsx
PC98: @matthewbrown77 I think you missed this, but apart from that LGTM

PC98 (approved):

PR #316, opened 2020-10-21
Add sorting support for filtered events

### LINKED ISSUE #293: [VMS] Add sorting support to `getByFilteredEventsCardData()`
* If you look at the Figma designs, the EventsListings page supports sort by "most signed up" and "closest to you".
* The default value is "most signed up".  Think about how you might want to modify `getByFilteredEventsCardData()`'s arguments to support these two sort types and modify the query accordingly.
* Do we want to add two new parameters to `getByFilteredEventsCardData()`, `lat` and `long`? If `lat` and `long` are `null` we can assume sort by "most signed up", else assume sort by "closest to you"? Think of different approaches.
* **Note:** currently, none of our events have volunteers, so it might be difficult to test sort by "most signed up". Message me and I can help you with some seed script.

matthebrown:
Closes #293

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -137,7 +137,7 @@ export async function getByFilteredEventsCardData({
   let findQuery = await createFilter({ causes, cities, times });
   let sortQuery = {};
 
-  if (lat === INVALID_COORDINATE && long === INVALID_COORDINATE) {
+  if (lat !== INVALID_COORDINATE && long !== INVALID_COORDINATE) {
     // We can't sort by distance because of $geoWithin, but we can add a $near component
     // to the query using the user's location ($near returns sorted results).
     findQuery = {
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -28,6 +28,7 @@ const CARD_FIELDS: Record<keyof EventCardDataType, 1> = {
 };
 const MILLISECONDS_IN_WEEK = 7 * 24 * 60 * 60 * 1000;
 const NEAREST_EVENTS_RADIUS = 20 / 3959; // radius for nearest events in radians (20 miles / earth's radius)
+const INVALID_COORDINATE = -999;
 
 export async function getUpcomingEventsCardData({
   date,
@@ -126,8 +127,8 @@ export async function getByFilteredEventsCardData({
   cities,
   times,
   page,
-  lat = 0,
-  long = 0,
+  lat = INVALID_COORDINATE,
+  long = INVALID_COORDINATE,
   totalCount
 }: FilterPageInformation) {
   const CARDS_PER_PAGE = 16;
@@ -136,7 +137,9 @@ export async function getByFilteredEventsCardData({
   let findQuery = await createFilter({ causes, cities, times });
   let sortQuery = {};
 
-  if (lat && long) {
+  if (lat === INVALID_COORDINATE && long === INVALID_COORDINATE) {
+    // We can't sort by distance because of $geoWithin, but we can add a $near component
+    // to the query using the user's location ($near returns sorted results).
     findQuery = {
       ...findQuery,
       "address.location": {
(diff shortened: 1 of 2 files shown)


matthebrown:
Here you can see the results of the same query with and without the location parameters. The events are properly reordered based on how close they are to the passed coordinates.
unsorted
sorted

Here is the result of the same query without location parameters. I updated the first two events to have 2 and 1 volunteer ids respectively (I removed them after testing the query).
volunteers

server/actions/events.ts
```diff
@@ -126,15 +127,35 @@ export async function getByFilteredEventsCardData({
...
+  lat = 0,
+  long = 0,
   totalCount
 }: FilterPageInformation) {
   const CARDS_PER_PAGE = 16;
   await Mongo();
 
-  const findQuery = await createFilter({ causes, cities, times });
+  let findQuery = await createFilter({ causes, cities, times });
+  let sortQuery = {};
```
matthebrown, line 138: Sorting with an empty object causes no changes to the way results are ordered.

server/actions/events.ts
```diff
@@ -126,15 +127,35 @@ export async function getByFilteredEventsCardData({
...
+
+  if (lat && long) {
+    findQuery = {
```
matthebrown, line 143: Due to the way that $geoWithin works, it is very complex to sort by distance after the query has completed. Instead, we can add another component to the query that uses $near to order the events by proximity to the user's location.

server/actions/events.ts
```diff
@@ -126,15 +126,35 @@ export async function getByFilteredEventsCardData({
   cities,
   times,
   page,
+  lat = 0,
+  long = 0,
```
PC98: Let's try to think of different values for `lat`, `long` here since 0, 0 is a valid coordinate that we could potentially be searching for. 

We can perhaps make a variable, `const INVALID_COORDINATE = -999;` or make `lat` and `long` `null`able (or `undefined`) to indicate that we want to sort by volunteers.

server/actions/events.ts
```diff
@@ -126,15 +126,35 @@ export async function getByFilteredEventsCardData({
...
   totalCount
 }: FilterPageInformation) {
   const CARDS_PER_PAGE = 16;
   await Mongo();
 
-  const findQuery = await createFilter({ causes, cities, times });
+  let findQuery = await createFilter({ causes, cities, times });
+  let sortQuery = {};
+
+  if (lat && long) {
```
PC98: This condition will have to change based on my above comment.

server/actions/events.ts
PC98: Might be helpful to explain this as a comment -- why we are doing this as part of `findQuery` as opposed to `sortQuery`. Also briefly mention that `$near` returns sorted results.

server/actions/events.ts
```diff
@@ -126,15 +126,35 @@ export async function getByFilteredEventsCardData({
...
+        $near: {
+          $geometry: {
+            type: "Point",
+            coordinates: [long, lat]
+          }
+        }
+      }
+    };
+  } else {
+    sortQuery = { volunteers: -1 };
```
PC98: nit: if we want to follow `findQuery` convention, `{ ... sortQuery , volunteers: -1 };`

PC98 (approved):
Thanks for testing this! You have also brought up a good point - we need to add volunteers to some of our events via a migration script. I will make a task for this.

Feel free to merge after making the minor change to default values.

server/actions/events.ts
PC98: `lat !== INVALID_COORDINATE && long !== INVALID_COORDINATE` ?

PC98 (commented):
i think you have an error in the conditional

matthebrown:
Good catch :sweat_smile:

PR #322, opened 2020-10-23
Add endpoint for filtered events

### LINKED ISSUE #296: [VMS] API endpoint and request method for filtered events server action
* After #293 is completed, let's make an API endpoint and a fetch request method for this server action.
* The API endponit should use `handleGetRequestWithPayloadResponse`. You may have some TS problems with specifying the `convertQueryToServerActionInput` function argument that `handleGetRequestWithPayloadResponse` expects. DM me and we can work together on this.
* Don't forget to make changes to `config.ts`.

matthebrown:
Closes #296

### DIFF
--- a/pages/api/getFilteredEvents.ts
+++ b/pages/api/getFilteredEvents.ts
@@ -1,0 +1,79 @@
+import { NextApiRequest, NextApiResponse } from "next";
+
+import {
+  getFilteredEventsCardData,
+  INVALID_COORDINATE
+} from "server/actions/events";
+import { getFilterValuesInQuery, FilterType, FilterValue } from "utils/filters";
+import { handleGetRequestWithPayloadResponse } from "utils/util";
+
+// @route   GET filtered events card data
+// @desc    Gets event data given filter parameters
+// @access  Public
+export default async (
+  req: NextApiRequest,
+  res: NextApiResponse
+): Promise<void> =>
+  handleGetRequestWithPayloadResponse(
+    req,
+    res,
+    getFilteredEventsCardData,
+    ["causes", "cities", "times", "lat", "long", "page", "totalCount", "date"],
+    queryRecord => {
+      const causes = getFilterValuesInQuery(
+        queryRecord,
+        "causes" as FilterType
+      ) as FilterValue<"cause">[];
+
+      const cities = getFilterValuesInQuery(
+        queryRecord,
+        "locations" as FilterType
+      );
+
+      const times = getFilterValuesInQuery(
+        queryRecord,
+        "times" as FilterType
+      ) as FilterValue<"time">[];
+
+      let lat;
+      if (queryRecord.lat === "") {
+        lat = INVALID_COORDINATE;
+      } else {
+        lat = Number(queryRecord.lat);
+      }
+
+      let long;
+      if (queryRecord.long === "") {
+        long = INVALID_COORDINATE;
+      } else {
+        long = Number(queryRecord.long);
+      }
+
+      const page = Number(queryRecord.page);
+      const totalCount = Number(queryRecord.totalCount);
+      const date = queryRecord.date;
+
+      if (
+        isNaN(lat) ||
+        isNaN(long) ||
+        isNaN(page) ||
+        isNaN(totalCount) ||
+        Array.isArray(date)
+      ) {
+        throw new Error(
+          "API call to getFilteredEvents did not receive data of the expected type!"
+        );
+      }
+
+      return {
+        causes,
+        cities,
+        times,
+        page,
+        lat,
+        long,
+        totalCount,
+        date
+      };
+    }
+  );
--- a/pages/api/getFilteredEvents.ts
+++ b/pages/api/getFilteredEvents.ts
@@ -1,10 +1,7 @@
 import { NextApiRequest, NextApiResponse } from "next";
 
-import {
-  getFilteredEventsCardData,
-  INVALID_COORDINATE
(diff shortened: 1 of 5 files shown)


matthebrown:
Sample output from postman
pm

pages/api/getFilteredEvents.ts
```diff
@@ -0,0 +1,79 @@
...
+  req: NextApiRequest,
+  res: NextApiResponse
+): Promise<void> =>
+  handleGetRequestWithPayloadResponse(
+    req,
+    res,
+    getFilteredEventsCardData,
+    ["causes", "cities", "times", "lat", "long", "page", "totalCount", "date"],
+    queryRecord => {
+      const causes = getFilterValuesInQuery(
```
matthebrown: I left the casting here for now since I wasn't sure exactly how to accomplish your suggestion for refactoring `PaginateWithFilter`, but we can talk about it before or after our Saturday meeting.

server/actions/events.ts
```diff
@@ -28,7 +29,7 @@ const CARD_FIELDS: Record<keyof EventCardDataType, 1> = {
 };
 const MILLISECONDS_IN_WEEK = 7 * 24 * 60 * 60 * 1000;
 const NEAREST_EVENTS_RADIUS = 20 / 3959; // radius for nearest events in radians (20 miles / earth's radius)
-const INVALID_COORDINATE = -999;
+export const INVALID_COORDINATE = -999;
```
matthebrown: There may be a better place for this since we are now using it in other files. I just decided to leave it here and export it for now to avoid magic numbers.

pages/api/getFilteredEvents.ts
PC98: So the changes I was suggesting to `PaginateWithFilter` [here](https://github.com/GTBitsOfGood/gen-soln/issues/296#issuecomment-714799742) are optional, we can/should do them though to make `const filters` in `utils/filters.ts` consistent with the `PaginateWithFilter` type.

pages/api/getFilteredEvents.ts
```diff
@@ -0,0 +1,79 @@
...
+      ) as FilterValue<"cause">[];
+
+      const cities = getFilterValuesInQuery(
+        queryRecord,
+        "locations" as FilterType
+      );
+
+      const times = getFilterValuesInQuery(
+        queryRecord,
+        "times" as FilterType
```
PC98: Let's not do this since "times" is really not a `FilterType`. TS will tell you that `queryRecord` definitely contains "times". This applies to all calls to `getFilterValuesInQuery` as well, so you can just do:
```javascript
const { causes, cities, times, date } = queryRecord;
const lat = Number(queryRecord.lat);
const long = Number(queryRecord.long);
const page = Number(queryRecord.page);
const totalCount = Number(queryRecord.totalCount);
```
Then, as I described [here](https://github.com/GTBitsOfGood/gen-soln/issues/296#issuecomment-714629902):

> when we receive this request on our back-end, req.query will represent this data as { cities: "", causes: "ARTS", times: "TOMORROW,WEEK" /* rest * / }, so basically all the array fields have been converted to strings.

Using your `if` condition, you should ensure that `cities`, `causes`, and `times` are `string`s. Then on line 69 you can do:
```javascript
return {
  causes: causes === "" ? [] : causes.split(",") as FilterValue<"cause">[],
  cities: /* similar to above */,
  times: /* similar to above */,
  // rest
};
```

pages/api/getFilteredEvents.ts
```diff
@@ -0,0 +1,79 @@
...
+      ) as FilterValue<"time">[];
+
+      let lat;
+      if (queryRecord.lat === "") {
```
PC98: I don't think `queryRecord.lat` or `queryRecord.long` can ever equal `""`. Notice that in the `getFilteredEvents` fetch request, you take in an object of type `FilterPageRequest`. In that type, `lat` and `long` are required `number`s. So here, `queryRecord.lat` and `queryRecord.long` will be string representations of those `number`s.

You can thus avoid setting them to `INVALID_COORDINATE` and exporting it/importing the value in this file.

utils/types.ts
```diff
@@ -92,6 +92,8 @@ export type LocationPageRequest = Pick<PageInformation, "page"> &
 export type DatePaginatedEventCards = PaginatedEventCards & PaginateWithDate;
 export type DatePageRequest = Pick<PageInformation, "page"> & PaginateWithDate;
 
+export type FilterPaginatedEventCards = PaginatedEventCards &
```
PC98, line 95: We should perhaps also include `PaginateWithLocation` and `PaginateWithDate` here, just like `FilterPageRequest`.

pages/api/getFilteredEvents.ts
matthebrown: Yeah my solution seemed very hacky. This makes much more sense.

pages/api/getFilteredEvents.ts
matthebrown: I was assuming that we would not be passing a value for `lat `or `long `if the user refused to share their location, but if we just pass the invalidate coordinate value then this is no longer an issue.

PC98: > I was assuming that we would not be passing a value for lat or long if the user refused to share their location

Yes, that would be ideal, but since the type is `number` right now, we will always have to pass some value. We can change the type though, to get around this.

PC98 (approved):
💯

PR #324, opened 2020-10-24
Update cause card styling

### LINKED ISSUE #307: [VMS] Modify cause card UI
* The text in `EventsPageCauseCard` is not always centered :/ see this:

* Let's fix this! Also refer to Figma design spec for any padding values, etc.

migueldlr:
Closes #307 

This change centers text in the cause card
We also add the `isSmall` prop to handle the two sizes in the Figma, defaulting to the large size


PC98 (approved):
Awesome! Do you know when we will have to use the non-small variant though?

PR #335, opened 2020-10-25
created cache to make filtering more efficient for loc and causes

Fattimo:
We used a single variable to point to a single cache, and both locations and causes are stored as keys on this cache. There is potential for collision of keys, so should we make a second cache variable, one for location and one for causes?

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -209,7 +209,7 @@ export const createFilter = async ({
       nonprofitId: { $in: idsWithCauses.flat() }
     };
   }
-  console.log(idsWithCauses);
+
   if (cities.length) {
     findQuery = {
       ...findQuery,
@@ -297,10 +297,8 @@ export async function getAllEventIds(): Promise<string[]> {
 function getNonprofitIdsByCause(causes: FilterPageRequest["causes"]) {
   return Promise.all(
     causes.map(async cause => {
-      if (filterCache.has(cause)) {
-        const nonProfitsWithCause: string[] = filterCache.get<string[]>(
-          cause
-        ) as string[];
+      const nonProfitsWithCause = filterCache.get<string[]>(cause);
+      if (nonProfitsWithCause != null) {
         return nonProfitsWithCause;
       } else {
         const nonProfitsWithCause: string[] = await Nonprofit.distinct("_id", {
@@ -322,8 +320,8 @@ function getCityPolygonCoordinates(cities: string[]) {
 
   return Promise.all(
     cities.map(async city => {
-      if (filterCache.has(city)) {
-        const cachedBounds = filterCache.get<number[][][]>(city);
+      const cachedBounds = filterCache.get<number[][][]>(city);
+      if (cachedBounds != null) {
         return cachedBounds;
       } else {
         const geocode = await client.geocode({
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -1,4 +1,5 @@
 import { Client } from "@googlemaps/google-maps-services-js";
+import NodeCache from "node-cache";
 
 import config from "config";
 import Mongo from "server/index";
@@ -16,6 +17,8 @@ import {
   FilterPaginatedEventCards,
   FilterPageRequest
 } from "utils/types";
+
+const filterCache = new NodeCache();
 
 const CARD_FIELDS: Record<keyof EventCardDataType, 1> = {
   name: 1,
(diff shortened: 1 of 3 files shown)


Fattimo:
image

server/actions/events.ts
```diff
@@ -268,31 +287,39 @@ function getCityPolygonCoordinates(cities: string[]) {
...
-        [
-          [north, east],
-          [south, east],
-          [south, west],
-          [north, west],
-          [north, east] // duplicate the first one in order to create a closed loop for mongo
-        ]
-      ];
+      if (filterCache.has(city)) {
+        return filterCache.get(city);
```
PC98: To help TS infer that this `number[][][]`, do something like this:
```javascript
const cachedBounds = filterCache.get<number[][][]>(city);
if (cachedBounds != null) {
  return cachedBounds;
}
// rest of your code
```

You should also take a similar approach when retrieving non-profit IDs for a cause.

server/actions/events.ts
```diff
@@ -168,9 +171,25 @@ export const createFilter = async ({
...
+          cause
+        });
+        idsWithCause.push(...nonProfitsWithCause);
+        filterCache.set(
+          cause,
+          nonProfitsWithCause,
+          MILLISECONDS_IN_WEEK / 1000
+        );
+      }
+    }
```
PC98: While this will work, this code is not parallelized like `getCityPolygonCoordinates`.

We should define a new method, which will use `Promise.all`:
```javascript
function someFn(causes: FilterPageRequest["causes"]) {
  return Promise.all(
    causes.map(async cause => {
      // your code
    })
  );
}
```
Though keep in mind `someFn` will return a `Promise<string[][]>` type now, so we'll have to use `.flat()`.

Then, after line 172, you can also parallelize the promises returned by `someFn` and `getCityPolygonCoordinates`:
```javascript
const [someVar, bounds] = await Promise.all([someFn(causes), getCityPolygonCoordinates(cities)]);
if (someVar.length) {
  // modify findQuery to include nonprofitId: { $in: someVar.flat() }
}
if (bounds.length) {
  // modify findQuery to filter location based on bounds
}
// rest
```
And yeah just use better variable names than what I have here.

server/actions/events.ts
```diff
@@ -168,9 +171,25 @@ export const createFilter = async ({
...
-    });
+    const idsWithCause = [];
+    for (const cause of causes) {
+      if (filterCache.has(cause)) {
+        const nonProfitsWithCause: string[] = filterCache.get(
+          cause
+        ) as string[];
+        idsWithCause.push(...nonProfitsWithCause);
+      } else {
+        const nonProfitsWithCause = await Nonprofit.distinct("_id", {
```
PC98: Lets use `const nonProfitsWithCause: string[]` or `as string[]` here

PC98 (approved):
> how we should deal with the case of new nonprofits being added into the db and not being cached under their cause

I think because we have a timeout/time-to-live, this should be fine? Like when the time expires, the cache lookup will result in a miss and the Mongo query will have to run.

Fattimo:
image

server/actions/events.ts
```diff
@@ -196,19 +199,18 @@ export const createFilter = async ({
...
-      cause: { $in: causes }
-    });
-
     findQuery = {
       ...findQuery,
-      nonprofitId: { $in: idsWithCause }
+      nonprofitId: { $in: idsWithCauses.flat() }
     };
   }
+  console.log(idsWithCauses);
```
PC98: nit remove

server/actions/events.ts
```diff
@@ -292,36 +294,68 @@ export async function getAllEventIds(): Promise<string[]> {
...
-          [south, east],
-          [south, west],
-          [north, west],
-          [north, east] // duplicate the first one in order to create a closed loop for mongo
-        ]
-      ];
+      if (filterCache.has(city)) {
+        const cachedBounds = filterCache.get<number[][][]>(city);
+        return cachedBounds;
+      } else {
```
PC98: Problem here is that TS will infer `cachedBounds` to be of type `number[][][] | undefined`, but we want it to be just `number[][][]`. Try this:

```javascript
const cachedBounds = filterCache.get<number[][][]>(city);
if (cachedBounds != null) {
  return cachedBounds;
} else {
// rest of your code
}
```
You don't need the `.has()`, `.get()` returns `undefined` for keys not in cache.

server/actions/events.ts
```diff
@@ -292,36 +294,68 @@ export async function getAllEventIds(): Promise<string[]> {
...
 }
 
+function getNonprofitIdsByCause(causes: FilterPageRequest["causes"]) {
+  return Promise.all(
+    causes.map(async cause => {
+      if (filterCache.has(cause)) {
+        const nonProfitsWithCause: string[] = filterCache.get<string[]>(
+          cause
+        ) as string[];
+        return nonProfitsWithCause;
```
PC98: Same as above, instead of all this casting:
```javascript
const nonProfitsWithCause = filterCache.get<string[]>(cause);
if (nonProfitsWithCause != null) {
  return nonProfitsWithCause;
} else {
// rest of your code
}
```

PR #336, opened 2020-10-26
[DMS] Add address fields, allow custom addresses

### LINKED ISSUE #282: [DMS] Add address text fields to billing step

ML-Chen:
Here's what it looks like right now:
image
Basically you can type in the "Billing Address" box and if you click on one of the options, it'll autofill that field and the city, state, and country.
I modified LocationAutocompleteField.tsx a good amount to customize it for DMS, and ~~later I'll have to add a prop to enable those customizations for DMS so that in VMS it'll function the same as before.~~ done
Resolves #282

### DIFF
--- a/components/LocationAutocompleteInput.tsx
+++ b/components/LocationAutocompleteInput.tsx
@@ -38,7 +38,9 @@ export type PlaceType = google.maps.places.AutocompletePrediction;
 interface Props {
   locationType: string;
   parentCallback: (value: PlaceType | string) => void;
-  filterOptions?: (options: PlaceType[]) => PlaceType[];
+  filterOptions?: (
+    options: Array<PlaceType | string>
+  ) => Array<PlaceType | string>;
   label?: string;
   fullWidth?: boolean;
   defaultValue?: PlaceType | string | null;
@@ -61,7 +63,7 @@ const LocationAutocompleteInput: React.FC<Props> = ({
   const { textStyle, highlightedText, inputAdornmentRoot } = useStyles();
   const [value, setValue] = useState<PlaceType | string | null>(defaultValue);
   const [inputValue, setInputValue] = useState("");
-  const [options, setOptions] = useState<PlaceType[]>([]);
+  const [options, setOptions] = useState<Array<string | PlaceType>>([]);
   const autocompleteService = useRef<google.maps.places.AutocompleteService | null>(
     null
   );
@@ -112,7 +114,7 @@ const LocationAutocompleteInput: React.FC<Props> = ({
       return undefined;
     }
 
-    if (inputValue === "" && typeof value !== "string") {
+    if (inputValue === "") {
       setOptions(value ? [value] : []);
       return undefined;
     }
@@ -123,9 +125,8 @@ const LocationAutocompleteInput: React.FC<Props> = ({
       },
       (results?: PlaceType[]) => {
         if (active) {
-          let newOptions = [] as PlaceType[];
-          console.log(value);
-          if (value && typeof value !== "string") {
+          let newOptions: Array<PlaceType | string> = [];
+          if (value) {
             newOptions = [value];
           }
 
@@ -164,7 +165,7 @@ const LocationAutocompleteInput: React.FC<Props> = ({
         newValue: PlaceType | string | null
       ) => {
         if (newValue) {
-          if (typeof newValue !== "string") setOptions(s => [newValue, ...s]);
+          setOptions(s => [newValue, ...s]);
           parentCallback(newValue);
         }
 
@@ -195,29 +196,31 @@ const LocationAutocompleteInput: React.FC<Props> = ({
         />
       )}
       renderOption={option => {
-        const matches =
-          option.structured_formatting.main_text_matched_substrings;
-        const parts = parse(
-          option.structured_formatting.main_text,
-          matches.map(match => [match.offset, match.offset + match.length])
-        );
+        if (typeof option !== "string") {
+          const matches =
+            option.structured_formatting.main_text_matched_substrings;
+          const parts = parse(
+            option.structured_formatting.main_text,
+            matches.map(match => [match.offset, match.offset + match.length])
+          );
 
-        return (
-          <Grid container alignItems="center">
-            {parts.map(({ text, highlight }, index) => (
-              <CoreTypography
-                variant="caption"
-                key={index}
-                className={clsx(highlight && highlightedText)}
-              >
-                {text.replace(/ /g, "\u00a0")}
+          return (
+            <Grid container alignItems="center">
+              {parts.map(({ text, highlight }, index) => (
+                <CoreTypography
+                  variant="caption"
+                  key={index}
+                  className={clsx(highlight && highlightedText)}
+                >
+                  {text.replace(/ /g, "\u00a0")}
+                </CoreTypography>
+              ))}
+              <CoreTypography variant="caption">
+                , {option.structured_formatting.secondary_text}
               </CoreTypography>
-            ))}
-            <CoreTypography variant="caption">
-              , {option.structured_formatting.secondary_text}
-            </CoreTypography>
-          </Grid>
-        );
+            </Grid>
+          );
+        }
       }}
     />
   );
--- a/components/LocationAutocompleteInput.tsx
+++ b/components/LocationAutocompleteInput.tsx
@@ -147,7 +147,9 @@ const LocationAutocompleteInput: React.FC<Props> = ({
   return (
     <Autocomplete
       getOptionLabel={option =>
-        typeof option === "string" ? option : option.description
+        typeof option === "string"
+          ? option
+          : option.structured_formatting.main_text
       }
       classes={{ noOptions: textStyle }}
       blurOnSelect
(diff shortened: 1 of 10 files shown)


ML-Chen:
Regarding how the form jumps up a bit when you click on the button and don't have all the fields filled out, I think that may have to do with how the form is too tall relative to the card, as when I remove the last line of form fields, it doesn't have that issue. @PC98  Do you know what code to change to make the card taller?

ML-Chen:
I was trying to use TypeScript generics so that LocationAutocompleteInput can use `PlaceType | string` if you use it allowing free-form text, and `PlaceType` if you don't. Unfortunately, it doesn't seem that TypeScript can support using generics like that. You can't use booleans like `T extends true` to control the types, it seems; see https://stackoverflow.com/questions/51432121/typescript-conditional-types-and-using-a-boolean-parameter-to-control-the-retur. I tried doing `<T extends (PlaceType | string) | PlaceType = PlaceType`, but TypeScript doesn't let you assign values to generics because it ["could be instantiated with a different type of constraint"](https://stackoverflow.com/questions/56505560/could-be-instantiated-with-a-different-subtype-of-constraint-object). So I had to do some `as PlaceType` type assertions.

components/donation/form/DonationPageForm.tsx
```diff
@@ -142,22 +142,21 @@ const DonationPageForm: React.FC<Props> = ({
...
             billingStep.email,
             billingStep.zipcode,
-            city,
-            billingStep.address.structured_formatting.main_text,
-            state,
-            country.slice(0, -1)
+            billingStep.city,
+            billingStep.addressLine,
+            billingStep.state,
+            billingStep.country.slice(0, -1)
```
PC98, line 170: Why the `.slice(0, -1)`?

components/donation/form/DonationPageForm.tsx
```diff
@@ -142,22 +142,21 @@ const DonationPageForm: React.FC<Props> = ({
...
-            country
-          ] = billingStep.address.structured_formatting.secondary_text.split(
-            ", "
-          );
+        if (
+          isPaymentStep &&
+          billingStep.addressLine &&
+          billingStep.city &&
+          billingStep.state &&
+          billingStep.country
```
PC98, line 161: Do we need these conditions on `billingStep`, given that `DonationPageFormBillingStep.tsx` marks all these inputs as required?

server/migrations/20201018194031-migrate-events-address-format.ts
```diff
@@ -28,7 +28,7 @@ export const up: MigrationFunction = async (db: Db) => {
           $set: {
             "address.text": {
               main: faker.random.arrayElement(MAIN_TEXT_OPTIONS),
-              // @ts-ignore: address.text is of type string
+              // @ts-ignore: addressLine.text is of type string
```
PC98, line 31: I don't think this change is related to your PR? Also ideally we should not modify already committed migration scripts

components/donation/form/reducer.ts
```diff
@@ -13,7 +13,11 @@ export interface BillingStepProps {
   firstName: string;
   lastName: string;
   email: string;
-  address: PlaceType | null;
+  addressField: PlaceType | string | null;
```
PC98, line 16: If we are okay with `freeSolo`, why do we need to store `addressField` in our redux store? Doesn't `addressLine` and other address fields suffice?

components/LocationAutocompleteInput.tsx
```diff
@@ -37,14 +37,19 @@ export type PlaceType = google.maps.places.AutocompletePrediction;
 
 interface Props {
   locationType: string;
-  parentCallback: (value: PlaceType) => void;
-  filterOptions?: (options: PlaceType[]) => PlaceType[];
+  parentCallback: (value: PlaceType | string) => void;
```
PC98, line 40: Instead of typing `PlaceType | string` over and over again, make a new type `type LocationValue = PlaceType | string;` or whatever you wanna call it and use that.

components/LocationAutocompleteInput.tsx
```diff
@@ -190,34 +202,36 @@ const LocationAutocompleteInput: React.FC<Props> = ({
                 <SearchIcon fontSize="inherit" />
               </InputAdornment>
             ),
-            classes: { root: textStyle }
+            ...(small && { classes: { root: textStyle } })
```
PC98, line 205: Is DMS not supposed to use caption typography style for their autocomplete? If not, the easiest thing to do might be to take text variant as a style:
```javascript
// Outside the component:
interface StyleProps {
  textVariant: keyof typeof typographyStyles;
}

const useStyles = makeStyles({
  // @ts-ignore: Pretty sure bug in mui types
  textStyle: (props: StyleProps) => typographyStyles[props.textVariant],
  // rest
});

interface Props extends StyleProps {
  // same expect remove "small"
}
// Inside the component:
const { textStyle, highlightedText, inputAdornmentRoot } = useStyles({
  textVariant
});
```

Then you can remove changes made on this line.

components/LocationAutocompleteInput.tsx
```diff
@@ -190,34 +202,36 @@ const LocationAutocompleteInput: React.FC<Props> = ({
...
-        return (
-          <Grid container alignItems="center">
-            {parts.map(({ text, highlight }, index) => (
-              <CoreTypography
-                variant="caption"
-                key={index}
-                className={clsx(highlight && highlightedText)}
-              >
-                {text.replace(/ /g, "\u00a0")}
+        if (typeof option !== "string") {
```
PC98, line 210: I see we are comparing against `"string"` type in a lot of places. Let's make a type guard outside of this component:
```javascript
const isAutocompletePrediction = (
  value: LocationValue | null
): value is PlaceType => value != null && typeof value !== "string";
```
Then simply call it everywhere you want to check `value`'s type: `isAutocompletePrediction(value)`

components/LocationAutocompleteInput.tsx
```diff
@@ -37,14 +37,19 @@ export type PlaceType = google.maps.places.AutocompletePrediction;
...
+    options: Array<PlaceType | string>
+  ) => Array<PlaceType | string>;
   label?: string;
   fullWidth?: boolean;
-  defaultValue?: PlaceType | null;
+  defaultValue?: PlaceType | string | null;
   required?: boolean;
   placeholder?: string;
   clearInputOnClose?: boolean;
+  freeInput?: boolean;
```
PC98, line 50: nit: use same name as prop, `freeSolo`

components/LocationAutocompleteInput.tsx
```diff
@@ -56,12 +61,16 @@ const LocationAutocompleteInput: React.FC<Props> = ({
...
-  clearInputOnClose = false
+  clearInputOnClose = false,
+  freeInput = false,
+  outlined = true,
+  small = true
 }) => {
   const { textStyle, highlightedText, inputAdornmentRoot } = useStyles();
-  const [value, setValue] = useState<PlaceType | null>(defaultValue);
+  // may be string if freeInput is true and user selects text not in autocomplete suggestions
+  const [value, setValue] = useState<PlaceType | string | null>(defaultValue);
```
PC98, line 71: just noticed this but we don't need the type specification `<PlaceType | string | null>` here.

components/LocationAutocompleteInput.tsx
```diff
@@ -56,12 +61,16 @@ const LocationAutocompleteInput: React.FC<Props> = ({
...
   const [inputValue, setInputValue] = useState("");
-  const [options, setOptions] = useState<PlaceType[]>([]);
+  const [options, setOptions] = useState<Array<string | PlaceType>>([]);
```
PC98, line 73: We should keep `options` as `PlaceType[]`, if we expect non-`freeSolo` to be used with autocomplete predictions. You might get a few TS errors, but they can be easily fixed using the type guard method suggested above. As an example:
```javascript
// in the useEffect
if (inputValue === "") {
  setOptions(isAutocompletePrediction(value) ? [value] : []);
  return undefined;
}
```

components/LocationAutocompleteInput.tsx
```diff
@@ -123,8 +132,7 @@ const LocationAutocompleteInput: React.FC<Props> = ({
       },
       (results?: PlaceType[]) => {
         if (active) {
-          let newOptions = [] as PlaceType[];
-
+          let newOptions: Array<PlaceType | string> = [];
```
PC98, line 135: Undo this change and use `isAutocompletePrediction` on line 136.

components/LocationAutocompleteInput.tsx
```diff
@@ -146,7 +154,9 @@ const LocationAutocompleteInput: React.FC<Props> = ({
   return (
     <Autocomplete
       getOptionLabel={option =>
-        typeof option === "string" ? option : option.description
+        typeof option === "string"
```
PC98, line 157: We will be able to use `isAutocompletePrediction` here as well.

components/donation/form/DonationPageFormBillingStep.tsx
```diff
@@ -126,17 +128,53 @@ const DonationPageFormBillingStep: React.FC<BillingStepProps> = ({
           locationType="address"
           fullWidth
           required
-          defaultValue={address}
+          defaultValue={addressLine}
           label="Billing Address"
+          freeInput={true}
```
PC98, line 133: nit: no need for `={true}` for `true` boolean props

components/donation/form/reducer.ts
```diff
@@ -13,7 +13,11 @@ export interface BillingStepProps {
...
+  addressLine: string | null;
```
PC98, line 17: Why not just `string`?

PC98 (commented):
If you wish to re-use the `isAutocompletePrediction` type guard and `LocationValue` type I have suggested in `reducer.ts` and `EventsPageLocationFilter.tsx` feel free to put them in some `utils/autocomplete.ts` file.

components/donation/form/DonationPageForm.tsx
ML-Chen: I have no idea. That was there before.

server/migrations/20201018194031-migrate-events-address-format.ts
ML-Chen: Yeah that was an accident, I'll revert it.

components/donation/form/DonationPageForm.tsx
KoolCards: Stripe needs a two letter country code (US instead of USA) which is why the slice is there. This won't work for other companies, but isn't an issue rn.

PC98: This might be problematic now though since I can enter any country in the country text box.

We should add a TODO/comment explaining this behavior and open a task to get the country code from Google's Place Details API.

PC98:
Closing this since changes went in as part of #371

PR #341, opened 2020-10-27
Show glimmer while waiting for user location

### LINKED ISSUE #290: [VMS] "Volunteer Events Near You" should show glimmering cards while fetching data
* Data for "Volunteer Events Near You" is fetched on the client-side. The user is first asked to share his location, then we kick-off a fetch request to the back-end, and upon receiving a response, we suddenly render a list of events when `nearestEvents.eventCards.length > 0` in `EventsPageMainContent.tsx`.
* This makes for a jerky UI experience. How about we show the "Volunteer Events Near You" list with glimmer cards **until the user grants permission to share location** _AND_ **the first set of cards have been fetched**? Essentially, we will want the UI to look like:

* Once data has been fetched, the glimmering cards can go away; if the fetched data returned no events, then we can unmount the component entirely.
* Also keep in mind that the `usePosition` hook in `EventsPageMainContent.tsx` will set `error` to a non-empty string when the user denies to share his location. Should we stop showing the glimmering cards then? 
* You may have to modify props/change functionality of the component added in #215.

migueldlr:
Closes #290 

Made `paginatedCardsData` optional in `CoreCardPaginationList.tsx` to show a glimmer when `paginatedCardsData` is null
Changed render condition in `EventsPageMainContent.tsx` to render as long as the user has not declined location permissions

### DIFF
--- a/components/core/list/CoreCardPaginationList.tsx
+++ b/components/core/list/CoreCardPaginationList.tsx
@@ -140,11 +140,18 @@ const CoreCardPaginationList = <CardData,>({
   // add an event listener and call the initial row size update
   useEffect(() => {
     window.addEventListener("resize", handleResize);
-    handleResize();
+
+    // we can't call handleResize directly here, the component might get
+    // unmounted immediately and we can't set state after unmount
+    const w = containerRef.current?.offsetWidth;
+    if (w != null) {
+      setRowSize(Math.max(1, Math.floor((w - MARGIN_ADJUSTMENT) / cardWidth)));
+    }
+
     return () => {
       window.removeEventListener("resize", handleResize);
     };
-  }, [handleResize]);
+  }, [cardWidth, handleResize]);
 
   // click handlers for buttons
   const nextPage = () => {
--- a/components/core/list/CoreCardPaginationList.tsx
+++ b/components/core/list/CoreCardPaginationList.tsx
@@ -44,11 +44,13 @@ const useStyles = makeStyles(({ palette }: Theme) =>
 );
 
 interface Props<CardData> {
-  paginatedCardsData?: PaginatedCards<CardData>;
+  paginatedCardsData: PaginatedCards<CardData>;
   fetchCards?: (newPage: number) => Promise<PaginatedCards<CardData>>;
   renderCard: (c: CardData) => JSX.Element;
   cardGlimmer: JSX.Element;
   cardWidth?: number;
+  shouldWait?: boolean;
+  setHasNoResults?: (b: boolean) => void;
 }
 
 const DEFAULT_ROW_SIZE = 4;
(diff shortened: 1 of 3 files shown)


components/events/EventsPageMainContent.tsx
```diff
@@ -87,26 +86,27 @@ const EventsPageMainContent: React.FC<Props> = ({ upcomingEvents }) => {
           </div>
         </>
       )}
-      {nearestEvents && nearestEvents.cards.length > 0 && (
```
PC98, line 90: I think now the "Volunteer Events Near You" text will still be rendered if `nearestEvents.cards.length == 0`. I am not sure if we want that, what do you think?

components/core/list/CoreCardPaginationList.tsx
```diff
@@ -63,20 +63,20 @@ const CoreCardPaginationList = <CardData,>({
 }: Props<CardData>) => {
   const classes = useStyles();
 
-  const [cards, setCards] = useState(paginatedCardsData.cards);
+  const [cards, setCards] = useState(paginatedCardsData?.cards ?? []);
```
PC98: One weird behavior is that the next button can still be clicked when this component is showing glimmers. It's possible that by the time the user grants location permission and the first set of cards are fetched from the back-end, he (for whatever reason) pressed the next button several times so `first` would be a large value and he won't see the first page of cards.

components/events/EventsPageMainContent.tsx
```diff
@@ -87,26 +86,27 @@ const EventsPageMainContent: React.FC<Props> = ({ upcomingEvents }) => {
...
-                  getNearestEvents({
-                    page,
-                    lat: nearestEvents.lat,
-                    long: nearestEvents.long
-                  })
-                }
-              />
-            </div>
+      {!hasPositionError && (
+        <div className={nearestEventsContainer}>
```
PC98, line 80: Another weird behavior - if you open the console's network tab, you will notice that two requests are being made to `/api/getNearestEvents` - one from this component and one from our pagination list component, for `page=0` and `page=1` respectively. But we should only fetch and display data for `page=0` unless the user clicks the next button.

PC98:
The only issue now seems to be:
```
react-dom.development.js?61bb:88 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in CoreCardPaginationList (at EventsPageEventList.tsx:27)
    in EventsPageEventList (at EventsPageMainContent.tsx:85)
    in div (at EventsPageMainContent.tsx:84)
    in div (at EventsPageMainContent.tsx:80)
```
This gets printed to the console when I visit the app on Chrome with location permission already set to blocked.

PC98 (approved):

PR #344, opened 2020-10-30
Fetch less information about a nonprofit from event page

### LINKED ISSUE #329: [VMS] Don't fetch entire non-profit object for the /events/id page
* The `getStaticProps` function for the `/events/id` page fetches the entire non-profit object.
* However, the EventDescription page only requires the non-profit name, id, and the about text.
* Write a new function that given a non-profit ID, only returns these fields. This function will be called instead of `getNonprofitById`.
* You may have to create a new type in `types.ts`; have a look at how `EventCardData` defines only a sub-fields of all possible event object fields and do something similar. 
* Don't forget to make relevant changes to the front-end code.

migueldlr:
Closes #329 

This PR adds the function `getNonprofitInfoForEventPageById()` to `server/actions/nonprofit.ts`, which returns only the name, about text, and ID of a nonprofit. This return type is `NonprofitInfoForEventPage`, a subtype of the `Nonprofit` interface. We change other usages in event page components accordingly.

### DIFF
--- a/server/actions/nonprofit.ts
+++ b/server/actions/nonprofit.ts
@@ -66,13 +66,25 @@ export async function getNonprofitById(_id: string): Promise<NonprofitType> {
   return nonprofit;
 }
 
+const NONPROFIT_FIELDS_FOR_EVENT_PAGE: Record<
+  keyof NonprofitInfoForEventPage,
+  1
+> = {
+  name: 1,
+  about: 1,
+  _id: 1
+};
+
 export async function getNonprofitInfoForEventPageById(
   _id: string
 ): Promise<NonprofitInfoForEventPage> {
   await Mongo();
 
   // Exclude donation information for now:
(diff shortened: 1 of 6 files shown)


server/actions/nonprofit.ts
```diff
@@ -63,6 +66,25 @@ export async function getNonprofitById(_id: string): Promise<NonprofitType> {
...
+    throw new Error(errors.nonprofit.INVALID_ID);
+  }
+
+  if (Array.isArray(nonprofit)) {
+    throw new Error(errors.GENERIC_TEXT);
+  }
+
+  // @ts-ignore: Temporary, until our Nonprofit Mongoose model is typed
+  return { name: nonprofit.name, _id: nonprofit._id, about: nonprofit.about };
+}
```
PC98: Consider implementing this function like how `getAllEventsCardData()` fetches only some data about an `Event` using `CARD_FIELDS` (defined in `server/actions/events.ts`).

PC98 (approved):

Issue #327, opened 2020-10-24
[VMS] `getNearestEventsCardData()` should return events sorted by distance


PC98:
* Currently, the Mongoose query in `getNearestEventsCardData()` uses `geoWithin` but that doesn't sort results by distance, see:  https://docs.mongodb.com/manual/reference/operator/query/geoWithin/#unsorted-results
* Try to figure out we can get sorted results, without using `.sort()`. Hint: look into `$near`

PR #346, opened 2020-10-30
added date support to getNearestEventCardData().

### LINKED ISSUE #328: [VMS] Ensure that `getNearestEventsCardData()` doesn't return past events
* If an event has a `startDate` < the current date, `getNearestEventsCardData()` should not return it.
* Try to think about how we can achieve this; keep in mind that we will want to use a static date and ensure that our pagination API calls don't break after this change.
* Have a look at how the Upcoming Events list does this. Relevant things to look at: `getUpcomingEventsCardData()` in `server/actions/events.ts`, the request in `requests/events.ts`, the API file in `pages/api` and the first request made in `pages/events/index.tsx`.

Fattimo:
changed typing of param of method to include date, changed api call to support a date

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -96,6 +96,7 @@ export async function getUpcomingEventsCardData({
 export async function getNearestEventsCardData({
   lat,
   long,
+  date,
   page
 }: LocationPageRequest): Promise<LocationPaginatedEventCards> {
   const CARDS_PER_PAGE = 4;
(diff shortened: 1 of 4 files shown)


server/actions/events.ts
```diff
@@ -107,10 +108,14 @@ export async function getNearestEventsCardData({
...
           $centerSphere: [[long, lat], NEAREST_EVENTS_RADIUS]
         }
+      },
+      startDate: {
+        $gte: new Date(date)
       }
     },
     CARD_FIELDS
   )
+    .sort({ startDate: 1 })
```
PC98: nit: remove this, we will want to sort by distance as part of #327

PC98 (approved):

PR #347, opened 2020-10-30
changed geoWithin to nearSphere for sorted nearest events results

### LINKED ISSUE #327: [VMS] `getNearestEventsCardData()` should return events sorted by distance

jade-law:
Closes #327 
Used the same lat, long and radius as the $geoWithin

### DIFF
--- a/server/actions/events.ts
+++ b/server/actions/events.ts
@@ -104,8 +104,12 @@ export async function getNearestEventsCardData({
   const result = await Event.find(
     {
       "address.location": {
-        $geoWithin: {
-          $centerSphere: [[long, lat], NEAREST_EVENTS_RADIUS]
+        $nearSphere: {
+          $geometry: {
+            type: "Point",
+            coordinates: [long, lat]
+          },
+          $maxDistance: NEAREST_EVENTS_RADIUS
         }
       }
     },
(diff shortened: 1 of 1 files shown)


server/actions/events.ts
```diff
@@ -104,8 +104,12 @@ export async function getNearestEventsCardData({
...
     {
       "address.location": {
-        $geoWithin: {
-          $centerSphere: [[long, lat], NEAREST_EVENTS_RADIUS]
+        $nearSphere: {
+          $geometry: {
+            type: "Point",
+            coordinates: [long, lat]
+          },
+          $maxDistance: NEAREST_EVENTS_RADIUS
```
PC98: Could you make sure that `NEAREST_EVENTS_RADIUS` is updated? Right now it's in radians, but `$nearSphere` expects the value to be in meters. Something like:
```javascript
const METERES_IN_A_MILE = 1609.34;
const NEAREST_EVENTS_RADIUS_IN_MILES = 20;
```
and then multiply the two here

PC98 (approved):

PR #354, opened 2020-11-03
Add next auth login support for admin

### LINKED ISSUE #225: next-auth backend setup
https://github.com/GTBitsOfGood/nextjs-sso-starter/

ayush-goyal:
Closes #225 

Adds Auth0 login functionality and automatically updates admins schema with user login info


package.json
```diff
@@ -33,10 +33,11 @@
     "isomorphic-unfetch": "^3.0.0",
     "js-cookie": "^2.2.1",
     "jsonwebtoken": "^8.5.1",
+    "mongodb": "^3.6.2",
     "mongoose": "^5.9.1",
     "nanoid": "^3.0.2",
     "next": "9.5.4",
-    "next-auth": "^3.1.0",
+    "next-auth": "GTBitsOfGood/next-auth#67498cf8fe11ca6b54cc7d4d62aefa6f0ac9992a",
```
PC98, line 40: out of curiosity, why do we need this change?

config.ts
```diff
@@ -28,6 +28,11 @@ export default {
     event: (path = "[id]"): string => `/events/${path}`,
     nonprofit: (path = "[id]"): string => `/nonprofits/${path}`
   },
+  auth0: {
+    clientId: process.env.AUTH0_CLIENT_ID,
+    clientSecret: process.env.AUTH0_CLIENT_SECRET,
+    domain: process.env.AUTH0_DOMAIN
```
PC98, line 34: if these are all secrets, add them to `.env.local.example` (**not the values**, just the names). If any of these env variables have good default values for the development environment, add the names and values to `.env`

package.json
```diff
@@ -102,4 +103,4 @@
     "utils": "utils",
     "config": "config.ts"
   }
-}
+}
```
PC98: nit: newline please

server/auth/BitsAuth0Provider.ts
```diff
@@ -0,0 +1,23 @@
...
+import config from "config";
+
+const BitsAuth0Provider = Providers.Auth0({
+  clientId: config.auth0.clientId as string,
+  clientSecret: config.auth0.clientSecret as string,
+  domain: config.auth0.domain as string
+});
+
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+BitsAuth0Provider.profile = (auth0Profile: any): any => {
```
PC98: remove `any` return type, you don't need to specify a return type

server/auth/AdminTypeORM.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * This class has an "id" field but because TypeORM is not
+ * really designed for MongoDB (NOSQL; "_id"), we can't include it as part of the type.
+ * NOTE: You can access the "id" field on this class, but if we include it in the
+ * type, a NULL "id" field will be created for all users
+ */
+export class Admin {
```
PC98, line 7: How does this relate to our existing admin schema? Do the schemas/models need to be merged somehow, does one have to be deleted?

pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,51 @@
...
+    models: {
+      User: AdminTypeORM
+    }
+  }),
+  database: config.db.url,
+  events: {},
+  callbacks: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    session: (session: any, user: Admin): Promise<AuthSession> => {
+      // See comment in @link /server/models/User.ts for why the User class does not include an "id" attribute
```
Comment (PC98) [inline pages/api/auth/[...nextauth].ts]: does this comment apply to our codebase?

pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,51 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment
+      const newUser: any = {
```
Comment (PC98) [inline pages/api/auth/[...nextauth].ts]: you shouldn't need the `: any` here

pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,51 @@
...
+  See src for correct definition: https://github.com/nextauthjs/next-auth/blob/main/src/adapters/typeorm/index.js */
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
+  adapter: (Adapters.TypeORM.Adapter as any)(config.db.url, {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
+    models: {
+      User: AdminTypeORM
+    }
+  }),
+  database: config.db.url,
+  events: {},
```
Comment (PC98) [inline pages/api/auth/[...nextauth].ts]: can this be skipped if it's empty?

pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,51 @@
...
+  /* The typescript definition for this function is wrong:
+  See src for correct definition: https://github.com/nextauthjs/next-auth/blob/main/src/adapters/typeorm/index.js */
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
+  adapter: (Adapters.TypeORM.Adapter as any)(config.db.url, {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
+    models: {
+      User: AdminTypeORM
+    }
+  }),
+  database: config.db.url,
```
Comment (PC98) [inline pages/api/auth/[...nextauth].ts]: > If the `adapter` option is specified it overrides the `database` option, only specify one or the other.

server/auth/AdminTypeORM.ts
PC98: And what about the user schema, like we have a `server/models/user.js` as well. Do the admin and user models need to be merged?

server/auth/AdminTypeORM.ts
```diff
@@ -0,0 +1,78 @@
...
+      type: "timestamp",
+      updateDate: true
+    }
+  }
+};
+
+export const AdminTypeORM = {
+  model: Admin,
+  schema: AdminSchema
+};
```
PC98: If you are exporting this, there shouldn't be the need to also `export` the `Admin` class separately. Also, consider making this a default export:
```javascript

export default {
  model: Admin,
  schema: AdminSchema
};
```

pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,51 @@
...
+      const newUser: any = {
+        ...session.user, // eslint-disable-line @typescript-eslint/no-unsafe-member-access
+        id: (user as any)["id"], // eslint-disable-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
+        firstName: user.firstName,
+        lastName: user.lastName,
+        email: user.email,
+        emailVerified: user.emailVerified,
+        image: user.image
+      };
+      return Promise.resolve({
```
Comment (PC98) [inline pages/api/auth/[...nextauth].ts]: you can avoid the `Promise.resolve` by making this function async.

pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,51 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
+    models: {
+      User: AdminTypeORM
+    }
+  }),
+  database: config.db.url,
+  events: {},
+  callbacks: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    session: (session: any, user: Admin): Promise<AuthSession> => {
```
Comment (PC98) [inline pages/api/auth/[...nextauth].ts]: `session: AuthSession` and then you won't have to use `as AuthSession` below.

pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,51 @@
...
+import NextAuth from "next-auth";
+import Adapters from "next-auth/adapters";
+
+import config from "config";
+import { Admin, AdminTypeORM } from "server/auth/AdminTypeORM";
+import BitsAuth0Provider from "server/auth/BitsAuth0Provider";
+
+export type AuthSession = {
+  user: Admin;
+  expiresAt: string;
```
Comment (PC98) [inline pages/api/auth/[...nextauth].ts]: `expires: string;` so that we are consistent with Next Auth's `SessionBase` type.

pages/api/auth/[...nextauth].ts
Comment (PC98) [reply pages/api/auth/[...nextauth].ts]: I think you can omit the return type.

package.json
ayush-goyal: Next auth is not very flexible, so to change the database adapter for our needs, we had to create our own fork. This is the same one that npp uses

pages/api/auth/[...nextauth].ts
Comment (ayush-goyal) [reply pages/api/auth/[...nextauth].ts]: session as the parameter is not of type AuthSession, we only return the type AuthSession

Comment (PC98) [reply pages/api/auth/[...nextauth].ts]: It seems to be of type `SessionBase`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/next-auth/index.d.ts#L114
However, since `SessionBase` is not exported, it seems NPP folks created their own interface, `AuthSession` which is very similar to `SessionBase`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/next-auth/_utils.d.ts#L26

You can perhaps omit the `: any` altogether, and import `InitOptions` from `"next-auth"` and do:
```javascript
const options: InitOptions = {
// rest
session: async (session, user: Admin): Promise<AuthSession> => {
// rest
} 
```

Comment (ayush-goyal) [reply pages/api/auth/[...nextauth].ts]: Ok cool thanks

server/auth/AdminTypeORM.ts
ayush-goyal: I'm exporting the Admin class separately so I can use it as a type, otherwise, it won't let me

server/auth/AdminTypeORM.ts
ayush-goyal: This is the admin schema that is defined for the typeorm adapter for next auth. The admins are the only ones who will login using oauth, so that's why its only defined for them. The users don't need to login so that's why I believe its separate.

PC98: I think we want both users and admins to be able to log-in. You can take care of that as a separate PR though. We should think about how this admin schema will interact with our mongoose schemas for admin and user. Should we somehow combine/share code between all three? Do we need separate mongoose schemas for admin and user?
 
@rithikgavvala

ayush-goyal: Oh interesting, yeah in that case we would have to combine the admin and user schema into one for mongo

PC98:
I forgot to ask but have you had the chance to test/try logging-in using a dummy user?

@rithikgavvala will let you take a final pass and approve. We will need some sort of follow-up task for https://github.com/GTBitsOfGood/gen-soln/pull/354#discussion_r516901687

ayush-goyal:
Yeah I logged in as a user through Auth0 and it works fine and updates the database

server/auth/AdminTypeORM.ts
rithikgavvala: We might need separate schemas based on how the project is structured now. But for now we have the functionality working for admin right?

PR #355, opened 2020-11-03
added default img for cause types

### LINKED ISSUE #350: [VMS] Update images for causes
* If you look at the Homepage designs on Figma, each cause has a different image. In this task, we will update our codebase to use the correct image for each cause instead of the default image.
* Firstly, modify `utils/filters.ts` so that "Arts" and "Culture and Humanities" are just one cause instead of two separate ones.
* Download images for each type of cause from Figma (should be 9 images in total) and put them in `public/` (feel free to create a new sub-folder in `public`).
* For each cause, you will have to specify the `imagePath` in `EventsPageCauseList.tsx`. You can be clever about how you name these image files to make this step easier.
* Lastly, delete the `public/defaultImages/defaultCause.png` since we won't have any use for it now.

Fattimo:


Fattimo:
image
Margins on images generally need to be tweaked, but they show up properly it seems

components/events/EventsPageCauseList.tsx
```diff
@@ -12,7 +12,7 @@ const CAUSE_CARDS: PaginatedCauseCards = {
   cards: filters["cause"].map(({ text, value }) => {
     return {
       cause: text,
-      imagePath: "/defaultImages/defaultCause.png",
+      imagePath: "/defaultImages/" + value + ".jpg",
```
PC98: nit: use a template string

PC98 (commented):
Could you play around with some CSS styles here: https://github.com/GTBitsOfGood/gen-soln/blob/develop/components/events/EventsPageCauseCard.tsx#L16 to ensure that the background image fits nicely in the container? Some CSS properties to look at: `backgroundSize`, `backgroundPosition`, etc. W3 schools is a good resource: https://www.w3schools.com/cssref/pr_background-position.asp

For the images, we can put them in something like `public/causes/` since these aren't default images but what we actually plan to use for each cause.

Fattimo:
image
text is hard to read on one of the cards, kind of a design thing though

PC98 (approved):
Ah I see... I will look into fixing this / contacting designers.

PR #370, opened 2020-11-06
Implement the filtered events list

### LINKED ISSUE #333: [VMS] Implement the filtered events list
* Now that #296 is complete, it's time to implement the filtered events list rendered in `EventsPageFilteredContent` with actual, real data from the back-end.
* Keep in mind that the data for the first page of filtered events will come from `getServerSideProps` in `events/pages/index.tsx`. Subsequent data will be fetched using the `getFilteredEvents` request defined in `requests/events.ts`.
* You may have to make some changes to the infinite scroll component added as part of #291, like adding props to it. Feel free to look at   `CoreCardPaginationList.tsx` to get an idea of what you can do.

matthebrown:
Closes #333


matthebrown:
So as of now, I have the page sorting properly. However, it still takes a few seconds for the position to propagate over the `InfiniteScroll` component. I couldn't remember exactly what the details we discussed to avoid passing `undefined` as the position, so I figured I'd just put a PR so we could discuss it.

components/events/EventsPageInfiniteScroll.tsx
```diff
@@ -19,50 +21,56 @@ const useStyles = makeStyles({
...
+        setCards(sortedEvents.cards);
+        setHasMore(!sortedEvents.isLastPage);
+        pageRef.current = sortedEvents.page;
+      })();
+    } else {
+      void (async () => {
+        const sortedEvents = await sortCards(-999, -999);
+        setCards(sortedEvents.cards);
+        setHasMore(!sortedEvents.isLastPage);
+        pageRef.current = sortedEvents.page;
```
matthebrown: There may be a way to reduce this duplicated code, but Typescript was complaining a lot about using a variable that hadn't been assigned so I left it be for now.

components/events/EventsPageInfiniteScroll.tsx
```diff
@@ -19,50 +21,56 @@ const useStyles = makeStyles({
...
+  const pageRef = useRef(filteredEvents.page);
+
+  useEffect(() => {
+    setCards(filteredEvents.cards);
+    setHasMore(!filteredEvents.isLastPage);
+    pageRef.current = filteredEvents.page;
+  }, [filteredEvents]);
+
+  useEffect(() => {
+    if (sort === "location" && position !== undefined) {
```
matthebrown: We also may want to make a task to go back and clean up these hardcoded references so we don't have to keep duplication these strings.

server/actions/events.ts
```diff
@@ -206,7 +207,7 @@ export const createFilter = async ({
   if (causes.length) {
     findQuery = {
       ...findQuery,
-      nonprofitId: { $in: idsWithCauses.flat() }
```
matthebrown, line 225: The method call to `flat()` was causing a type error. I replaced it with the lodash `flatten()` function for now until it could be troubleshot further.

pages/events/index.tsx
```diff
@@ -40,11 +45,12 @@ export const getServerSideProps = async (
   // Don't remove the async otherwise InferGetStaticPropsType won't work as expected
   // TODO: Use this eventually, if we need common props between filtered and unfiltered event pages.
   const commonProps = {};
-
+  const date = new Date().toJSON();
+  
   if (getFilterCountFromQuery(context.query) === 0) {
     const date = new Date();
```
PC98, line 51: remove line 51: `const date = new Date();` to fix the type error

PR #371, opened 2020-11-06
[DMS] Address autocomplete text fields

### LINKED ISSUE #282: [DMS] Add address text fields to billing step

KoolCards:
- Merges @ML-Chen work with mine to close #282 .


components/LocationAutocompleteInput.tsx
```diff
@@ -11,8 +11,13 @@ import { SearchIcon } from "@core/icons";
...
 import config from "config";
 
+interface StyleProps {
+  textVariant?: keyof typeof typographyStyles;
+}
+
 const useStyles = makeStyles({
-  textStyle: typographyStyles.caption,
+  // @ts-ignore: Pretty sure bug in mui types
+  textStyle: (props: StyleProps) => typographyStyles[props.textVariant], // eslint-disable-line
```
KoolCards: Couldn't figure out why I was getting a lint error here.

components/donation/form/DonationPageForm.tsx
```diff
@@ -36,6 +36,12 @@ const useStyles = makeStyles({
     marginTop: 8,
     marginBottom: 8,
     minHeight: 260
+  },
```
KoolCards, line 40: Increasing the card height only for the Billing Step because of the added text fields.

components/LocationAutocompleteInput.tsx
PC98: You are getting lint error since `textVariant` can be `undefined`. You should either make it required or provide a default value.

components/LocationAutocompleteInput.tsx
```diff
@@ -35,32 +40,43 @@ function loadScript(src: string, position: HTMLElement | null, id: string) {
...
+  outlined = true,
+  textVariant = "caption"
 }) => {
-  const { textStyle, highlightedText, inputAdornmentRoot } = useStyles();
+  const { textStyle, highlightedText, inputAdornmentRoot } = useStyles({
+    textVariant
+  });
   const [value, setValue] = useState<PlaceType | null>(defaultValue);
-  const [inputValue, setInputValue] = useState("");
+  const [inputValue, setInputValue] = useState(defaultInputValue);
```
PC98: I like the idea of using a `defaultInputValue`. We can consider removing `defaultValue` though since no component seems to be using it and `defaultValue ` vs. `defaultInputValue ` is confusing.

components/LocationAutocompleteInput.tsx
```diff
@@ -156,27 +173,30 @@ const LocationAutocompleteInput: React.FC<Props> = ({
...
         }
-
-        setValue(newValue);
       }}
       onInputChange={(event, newInputValue) => {
         setInputValue(newInputValue);
+        parentInputChangeCallback && parentInputChangeCallback(newInputValue);
       }}
-      onClose={() => void (clearInputOnClose && setValue(null))}
+      disableClearable={!clearInputOnClose}
```
PC98: Why do we need `disableClearable={!clearInputOnClose}`?

Also I think you might have to edit `onClose` to be `onClose={() => void (clearInputOnClose && setInputValue(''))}` since `inputValue` controls what gets displayed to the user.

components/LocationAutocompleteInput.tsx
```diff
@@ -35,32 +40,43 @@ function loadScript(src: string, position: HTMLElement | null, id: string) {
...
+  clearInputOnClose = false,
+  freeSolo = false,
+  outlined = true,
+  textVariant = "caption"
 }) => {
-  const { textStyle, highlightedText, inputAdornmentRoot } = useStyles();
+  const { textStyle, highlightedText, inputAdornmentRoot } = useStyles({
+    textVariant
+  });
   const [value, setValue] = useState<PlaceType | null>(defaultValue);
```
PC98: now that we are not using `value` to control the state of our auto-complete, do you think we can just use a `useRef` to store it as opposed to `useState`?

components/LocationAutocompleteInput.tsx
```diff
@@ -156,27 +173,30 @@ const LocationAutocompleteInput: React.FC<Props> = ({
       includeInputInList
       filterSelectedOptions
       fullWidth={fullWidth}
-      value={value}
+      value={inputValue}
       onChange={(
         event: React.ChangeEvent<unknown>,
-        newValue: PlaceType | null
+        newValue: PlaceType | string | null
       ) => {
```
PC98: I just realized we can omit these types:
```javascript
onChange={(event, newValue) => {
```

components/donation/form/DonationPageForm.tsx
```diff
@@ -36,6 +36,12 @@ const useStyles = makeStyles({
...
+  billingContentContainer: {
+    flex: 7,
+    marginTop: 8,
+    marginBottom: 8,
```
PC98: remove the duplicate styles and use `clsx` below (see suggestion)

components/donation/form/DonationPageForm.tsx
```diff
@@ -232,7 +239,7 @@ const DonationPageForm: React.FC<Props> = ({
           curStepIndex={curStepIndex}
           stepTitles={STEPS.map(_ => _.title)}
         />
-        <div className={contentContainer}>{componentJSX}</div>
+        <div className={dynamicContentContainer}>{componentJSX}</div>
```
PC98:
```javascript
<div
  className={clsx(
    contentContainer,
    step.title === "Billing" && billingContentContainer
  )}
>
```
I am also guilty of doing this but let's try to avoid using `useMemo` for inexpensive computation like string equality.

components/LocationAutocompleteInput.tsx
KoolCards: I added disableClearable because I personally didn't like the "x" button to delete the address input, and I don't think address inputs generally have this feature. Good catch on the second point.

ML-Chen:
When I go from to the Billing step, the header Donation Amount > Billing > Payment > Review jumps up a tiny but noticeable amount, and it jumps back down when I leave the step. Only at certain zoom levels though for some reason. The image in the background also shifts since the page has resized. Maybe we could try to just put the City, State, and Zipcode all on one line?

components/LocationAutocompleteInput.tsx
```diff
@@ -11,8 +11,13 @@ import { SearchIcon } from "@core/icons";
...
 import config from "config";
 
+interface StyleProps {
+  textVariant: keyof typeof typographyStyles;
+}
+
 const useStyles = makeStyles({
-  textStyle: typographyStyles.caption,
+  // @ts-ignore: Pretty sure bug in mui types
+  textStyle: (props: StyleProps) => typographyStyles[props.textVariant],
```
ML-Chen, line 20: I think Material UI expects the function to return a CreateCSSProperties object like `{ root: typographyStyles[props.textVariant] }`.
nah nvm this seems to work for some reason:
```ts
const useStyles2 = makeStyles({
  // style rule
  foo: (props: { color: "blue"; backgroundColor: "blue" }) => ({
    backgroundColor: props.backgroundColor
  }),
  bar: {
    // CSS property
    color: (props: { color: "blue"; backgroundColor: "blue" }) => props.color
  }
});
```

PR #373, opened 2020-11-06
support userId in donation schema

### LINKED ISSUE #367: [VMS] Modify donation schema to include user-id
* Let's establish a one-to-one relationship between donations and User IDs.
* However certain donations can have a null user ID.
* Make modifications to the `logDonation` server action and request to support user ID. Right now, the front-end code can pass null as the user ID but let's add a TODO saying that this should be changed.
* No need for a migration script since we don't have any existing donations in our DB.

Fattimo:
added userId in donation schema. changed donation typing in types. modified method logDonation to take a userId as well. corrected misspelling of donation


server/actions/donation.ts
```diff
@@ -11,11 +11,13 @@ export async function logDonation({
   name,
   email,
   amount,
+  userId,
   nonprofitId
 }: DonationType): Promise<void> {
   await Mongo();
 
   const nonprofit = await Nonprofit.findOne({ _id: nonprofitId });
+  const user = await Donation.findOne({ _id: userId }); // TODO: Don't allow front end to pass null resulting userId
```
PC98: Technically this should be `User.findOne` right? Also right now we haven't decided if all Donations require a user ID, so could you make necessary changes to the Mongoose schema and the types in `utils.ts` to support `null` user ID or make user ID optional?

PC98: If a user ID is specified and there is no user with that ID, we should throw an error, just like on line 22 we throw an error if the nonprofit doesn't exist.

server/actions/donation.ts
```diff
@@ -4,27 +4,35 @@ import Stripe from "stripe";
...
   email,
   amount,
+  userId,
   nonprofitId
-}: Omit<DonationType, "timestamp">): Promise<void> {
+}: LoggedDonation): Promise<void> {
   await Mongo();
 
   const nonprofit = await Nonprofit.findOne({ _id: nonprofitId });
+  const user = await User.findOne({ _id: userId }); // TODO: Don't allow front end to pass null resulting userId
```
PC98: A couple of things:
can we do this and line 20 in parallel, using `Promise.all`?

can we modify line 27 to **not** throw the error if `userId` equals the empty string (as you have done in `DonationPageForm.tsx`), just temprarily? Because otherwise, donations will fail when DMS team is testing their UI.

PC98 (approved):
A couple of minor things, but tbh Matt it's okay if you don't do this now; we won't need this task for this semester's demo

PR #376, opened 2020-11-07
Remove event card glimmer component, refactor event card

### LINKED ISSUE #331: [VMS] Reduce style duplication between card and card glimmer components
* We should figure out a way to share certain styles between `EventsPageEventCard` and `EventsPageEventCardGlimmer` component This will help DRYup our code and make it more maintainable. 
* Feel free to do this however you see fit - perhaps create a HOC, or combine the glimmer and the card into a single component that supports a prop to control if the glimmering version should be displayed or the normal one, or create some sort of style object that can be imported (see `typographyStyles` in `core/typography/CoreTypography.tsx` as an example), or any other approach you can think of.
* **No need to do this for the causes card.** The causes glimmer card will probably be removed since we are not really using it.

migueldlr:
Closes #331 

Remove `EventsPageEventCardGlimmer` and instead render a glimmer from `EventsPageEventCard` if `eventCardData` or `onClick` is `undefined`

Change usages in `EventsPageEventList` and `EventsPageInfiniteScroll`


components/events/EventsPageEventCard.tsx
```diff
@@ -62,46 +63,63 @@ const useStyles = makeStyles(({ palette }: Theme) =>
 );
 
 interface Props {
-  // see EventCardData under util/types
-  eventCardData: EventCardData;
-  onClick: () => void;
+  /* if either of these are undefined, we render a glimmer instead */
+  eventCardData?: EventCardData;
+  onClick?: () => void;
```
PC98: I think the idea of keeping everything in just one component, but ideally, we should make things more explicit. One way might be using TS's discriminated unions:
```javascript
type Props =
  | {
      eventCardData: EventCardData;
      onClick: () => void;
      type: "data";
    }
  | { type: "glimmer" };

const EventsPageEventCard: React.FC<Props> = props => {
  // anything you need
  let image;
  switch (props.type) {
    case "data":
      image = // whatever
      // assign other variables
      break;
    case "glimmer":
      image = // whatever
      // assign other variables
      break;
    default: {
      const _exhaustiveCheck: never = props;
      return _exhaustiveCheck;
    }
  }
  return // whatever
}
```

components/events/EventsPageEventList.tsx
```diff
@@ -27,7 +26,7 @@ const EventsPageEventList: React.FC<Props> = ({
     <CardPaginationList
       paginatedCardsData={paginatedEventCardsData}
       fetchCards={getMoreEvents}
-      cardGlimmer={<EventsPageEventCardGlimmer />}
+      cardGlimmer={<EventsPageEventCard />}
```
PC98: you will need to specify `type="glimmer"` as a prop

components/events/EventsPageInfiniteScroll.tsx
```diff
@@ -75,7 +74,7 @@ const EventsPageInfiniteScroll: React.FC = () => {
     >
       {Array.from({ length: CARDS_PER_PAGE }, (_, i) => (
         <Grid key={`glimmer_${i}`} item>
-          <EventsPageEventCardGlimmer />
+          <EventsPageEventCard />
```
PC98: same here, `type="glimmer"`

PC98: And you might need `type="data"` wherever the card is being supplied with event data.

PC98 (approved):

components/events/EventsPageInfiniteScroll.tsx
migueldlr: I'm ashamed I didn't catch these

PC98: haha no worries

PR #380, opened 2020-11-11
Added barebones nonprofit

Cullen101:


pages/nonprofits/[id].tsx
```diff
@@ -23,7 +24,7 @@ const NonprofitPage: NextPage<InferGetStaticPropsType<
     return <ErrorPage statusCode={404} />;
   }
 
-  return null;
+  return <NonprofitProfilePage {...props} />;
```
Comment (PC98) [inline pages/nonprofits/[id].tsx]: Should this be `<NonprofitPage {...props} />`?

server/migrations/20201111000721-remove-background.ts
```diff
@@ -0,0 +1,54 @@
+import { MigrationFunction } from "migrate-mongo";
+import { Db } from "mongodb";
```
PC98: I think this file shouldn't be part of this PR

PR #382, opened 2020-11-11
Cullen/348 remove bg image

Cullen101:


server/migrations/20201111004136-remove-background.ts
```diff
@@ -0,0 +1,53 @@
...
+    .collection("nonprofits")
+    .updateMany({}, { $unset: { background: "" } });
+};
+
+export const down: MigrationFunction = async (db: Db) => {
+  await db
+    .collection("nonprofits")
+    .updateMany(
+      { _id: { $in: nonprofits.map(_ => _._id) } },
+      { $set: { background: nonprofits.find(_id => _id)?.background } }
```
PC98: Let's do `down()` for all non-profits as well since technically the `up()` function also updates all non-profits. For each non-profit, we can assume the background imag was located at 
```javascript
`url(/backgrounds/${id}.png)` // this is a template string in JavaScript
```

PC98 (commented):
Feel free to also delete the `public/background` folder

Issue #369, opened 2020-11-05
[CORE][VMS] Component for non-profit card and corresponding glimmer card


PC98:
* Refer to the "Non-profits" page (select this on the left-hand side) of the Figma page for designs of the non-profit card.
* Create a component corresponding to this card. Model the props after the `EventsPageEventCard.tsx` component - have a `onClick` and a data prop. You might have to create a new type to represent this data, refer to how `EventCardData` type is defined.
* Note that while each non-profit will have **a logo image**, we will be using the same background image for each card (see #348 for details). Download the background image from Figma and save it somewhere in `public/`.
* Create a corresponding glimmer/loading card component. Feel free to look at how `EventsPageEventCardGlimmer` has been implemented.
* You will have to use some dummy data to test out the UI. For testing purposes, feel free to render this card as part of one of the event lists, like `EventsPageCauseList.tsx`.

matthebrown:
@kw143 Is there a time later this week that works to pair program on this? If not, we could each take a part of the ticket and work separately. I've started looking into the `NonprofitCard` if you're interested in taking the glimmer.

matthebrown:
@PC98 @kw143 I finished the nonprofit card and pushed the changes to `upstream/matthewbrown/non-profit-card-and-glimmer`. Feel free to work off that for the glimmer component. Gonna refrain from putting up a PR until both parts of the ticket are finished. 
![goodwill_card](https://user-images.githubusercontent.com/45107667/98747860-6874b180-2386-11eb-9cac-eb57f475c5ef.png)
![hearts_card](https://user-images.githubusercontent.com/45107667/98747864-6a3e7500-2386-11eb-846e-ce95a9deb34a.png)

PC98:
@matthewbrown77 This looks awesome!
Feel free to put out a PR though, it's okay if the glimmer isn't ready yet we can still merge your changes and someone else can work on it.

kw143:
Sorry about not responding earlier. I did not realize I was assigned this task again with a different person, and Andrew and I have been working on this separately. I just noticed it because in the pull request I got an error for duplicate nonprofitcarddata type. 

You can always message me on slack, I check it everyday. Sorry about missing this issue again. 
The same work is covered in the pull request I just made. And the cards are already connected to the carousel and with backend data, although it seems like you added the address and supporters, which I did not do, since there was no data to pull from.

kw143:
Will probably switched to use yours later tonight. 

Done

PR #383, opened 2020-11-11
Adds nonprofit card

matthebrown:
Closes part of #369


components/core/cards/NonprofitCard.tsx
```diff
@@ -0,0 +1,189 @@
...
+                  void router.push(
+                    config.pages.donate(),
+                    config.pages.donate(nonprofitCardData._id)
+                  );
+                }}
+              >
+                Donate
+              </CoreButton>
+              <div className={supporters}>
+                <CoreTypography variant="h4">5 Supporters</CoreTypography>
```
matthebrown, line 177: This will need to be modified once the nonprofits are updated with the `supporters` field.

matthebrown:
Not sure why the CI/CD check is failing. The lint runs without issue locally.

components/core/cards/NonprofitCard.tsx
```diff
@@ -0,0 +1,189 @@
...
+      <div className={cardContainer}>
+        <div className={card}>
+          <div className={header}>
+            <img className={logo} alt={`${nonprofitCardData.name}`} />
+            <div className={headerInfo}>
+              <CoreTypography variant="h2">
+                {nonprofitCardData.name}
+              </CoreTypography>
+              <CoreTypography variant="h5" className={location}>
+                Atlanta, GA
```
matthebrown, line 150: We will also need some type of `city` or `location` field to update this properly. We could try to pull info from the nonprofit's `events`, but a nonprofit could have events in multiple cities.

PC98: Ah good catch. I will ask Core if they are taking care of this.

components/core/cards/NonprofitCard.tsx
```diff
@@ -0,0 +1,189 @@
...
+        top: 0,
+        right: 0,
+        bottom: 0,
+        left: 0,
+        opacity: 0.8
+      }
+    },
+    logo: {
+      width: "124px",
+      height: "124px",
```
PC98: nit: just `124`

components/core/cards/NonprofitCard.tsx
```diff
@@ -0,0 +1,189 @@
...
+      marginLeft: 24,
+      backgroundColor: "white",
```
PC98: let's use `palette.background.paper` whenever you want to use "white" as a background color

components/core/cards/NonprofitCard.tsx
```diff
@@ -0,0 +1,189 @@
...
+      textTransform: "capitalize",
+      pointerEvents: "none"
+    },
+    aboutContainer: {
+      width: 478,
+      height: 54,
+      marginTop: 20,
+      marginLeft: 24,
+      fontSize: 14,
+      lineHeight: "18.2px",
```
PC98, line 97: might be able to drop the `px` and just a number, `18.2`

components/core/cards/NonprofitCard.tsx
```diff
@@ -0,0 +1,189 @@
...
+              <CoreTypography variant="h2">
+                {nonprofitCardData.name}
+              </CoreTypography>
+              <CoreTypography variant="h5" className={location}>
+                Atlanta, GA
+              </CoreTypography>
+              <CoreButton variant="contained" className={cause}>
+                <CoreTypography variant="caption">
+                  {nonprofitCardData.cause.toLowerCase()}
+                </CoreTypography>
```
PC98: Could you use a Chip instead: https://material-ui.com/components/chips/

components/core/cards/NonprofitCard.tsx
```diff
@@ -0,0 +1,189 @@
...
+      zIndex: 1
+    },
+    location: {
+      marginTop: 8
+    },
+    cause: {
+      height: 34,
+      marginTop: 12,
+      background: "white",
+      color: "#333333",
```
PC98: Is this gray-80? try using `palette.text.primary`
For `background: "white"`, try `backgroundColor: palette.background.paper` instead.

Though if we switch to using material-ui's Chip, you might not have to specify text color or background-color

utils/types.ts
```diff
@@ -67,6 +67,8 @@ export type CauseCardData = {
   filterValue: string;
 };
 
+export type NonprofitCardData = Omit<Nonprofit, "stripeAccount" | "events">;
```
PC98, line 70: Okay I think the reason you are getting a type error is that a change went into develop branch that removed background images from all non-profits, so line 124 in `NonprofitCard.tsx` will error out. 
You can use the same background image on all non-profit cards by downloading it from Figma.

PC98 (commented):
Awesome! Left some feedback but you don't have to prioritize working on it.
We will focus on getting the "Other Events Like This" list completed over the next few days.

components/core/cards/NonprofitCard.tsx
matthebrown: `lineHeight` rounds to a multiple, so we have to leave the "px"

PC98 (approved):

PR #384, opened 2020-11-11
do not render EventsPageFilteredHeaderSelect when there are 0 results

stevenkchen:
Added check to see if there were zero results, if there were 0 results, EventsPageFilteredHeaderSelect is not rendered.

### DIFF
--- a/components/events/header/EventsPageFilteredHeader.tsx
+++ b/components/events/header/EventsPageFilteredHeader.tsx
@@ -26,12 +26,20 @@ const EventsPageFilteredHeader: React.FC<Props> = ({
 }) => {
   const { header } = useStyles();
 
-  return (
-    <header className={header}>
-      <CoreTypography variant="h2">{resultsLength} results</CoreTypography>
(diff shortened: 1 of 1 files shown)


components/events/header/EventsPageFilteredHeader.tsx
```diff
@@ -26,12 +26,20 @@ const EventsPageFilteredHeader: React.FC<Props> = ({
 }) => {
   const { header } = useStyles();
 
-  return (
-    <header className={header}>
-      <CoreTypography variant="h2">{resultsLength} results</CoreTypography>
-      <EventsPageFilteredHeaderSelect setPosition={setPosition} />
-    </header>
-  );
```
PC98: to avoid duplicating code relted to `CoreTypography`, you could just do:
```javascript
return <header className={header}>
  <CoreTypography variant="h2">{resultsLength} results</CoreTypography>
  {resultsLength > 0 &&  <EventsPageFilteredHeaderSelect setPosition={setPosition} />} 
  </header>
```

PC98 (approved):

Issue #237, opened 2020-10-06
Building the support portion of the landing page UI


rithikgavvala:
![image.png](https://images.zenhubusercontent.com/5c410ae3a8dc8254c74193d0/329a0e48-3d36-4af9-a6f5-4411bd8c8f21)
The general steps to follow for this story would be as follows:
1. Create the UI for the cause and support. I will find the image assets for the various background pictures on the cards: Health, Religion, etc. 
2. Build the UI for the non-profit cards: ![image.png](https://images.zenhubusercontent.com/5c410ae3a8dc8254c74193d0/9aef98e0-6541-4116-ac4c-325a9930717f)
From there building a slider/pagination for to slide through the different non-profit cards.
3. Creating a function  'getNonProfitsByDate' in  `/server/actions/nonprofits` to get non-profits that have been recently added to the platform

PR #389, opened 2020-11-13
Landing page cause grid and nonprofit carousel

kw143:
This pull request closes the following issue: #237 

The issue was blocked for a while because of waiting on other blocking tasks and resources needed.
The issue also kept growing exponentially large as we dive into it, since there are many things connected to it that were not previously considered, such as incomplete backend schema.

The carousel of non-profit cards on the landing page is implemented together with the cause grid above it.

The api to request non-profit data, backend action to getNonprofitByDate was also included in this pull request.

There was also a data migration included in this pull request, and a schema change to non-profit for a new date attribute for requesting recently joined non-profits.

I encountered some issue with centering the carousel using the CoreCardPaginationList. The margin it has and the way it calculates rowSize does not work very well with centering alignment. I copied the component and modified the styling of the copy to center the carousel and maintain responsiveness (didn't modify the original because it might break other lists in the project). I changed the carousel width calculation and referencing the width w from window.innerWidth instead of using div ref because of a size change after center alignment.

Since the database currently does not have an address attribute, the cause list is empty for each non-profit, those data are not implemented to be shown on the non-profit card yet. (Edit: they are shown with placeholder now with Matthew's version)

The current descriptions of the non-profits also do not describe the non-profits very well. Using the about attribute of nonprofit schema from the database, it should be the same one as shown on the donation pages.

All components are done according to the Figma design and are ready to be merged.


kw143:
There are some conflicts with nonprofitcarddata since the card was implemented twice differently. Will fix it later tonight.

rithikgavvala (approved):

PR #396, opened 2020-11-15
Make landing page default route

aniruddhamurali:
Changed default route from donate page to landing page.


PC98 (commented):
Instead of pushing to `/home`, let's remove the `/home` route and have `pages/index.ts` render the `Home` component.

aniruddhamurali:
Does this mean I just have to `return Home` in `const IndexPage`? I'm a bit confused about how this works.

PC98:
Yep exactly.
You can delete the `pages/home.tsx` file and copy over relevant pieces of code (like the `return <Home` part)  from that file to `pages/index.ts` (please also rename `index.ts` to `index.tsx).

aniruddhamurali:
I'm still having some trouble with this - I think the main difficulty for me is passing in the props for the home page in `index.tsx`.

PC98:
I think you can literally just rename `pages/home.tsx` to `pages/index.tsx` and delete the existing `pages/index.ts` file

PC98:
Once you do that^ feel free to also delete the `getDefaultNonprofitId()` function from our codebase since it won't be used anymore

config.ts
```diff
@@ -21,6 +21,7 @@ export default {
   baseUrl: process.env.BASE_URL,
   pages: {
     index: "/",
+    home: "/home",
```
PC98: nit: delete

aniruddhamurali:
It's hilarious that this was as simple as renaming the file LOL. Thanks so much for the help!

PC98 (approved):

